PR #289, opened 2015-01-04
add <<-alias for add-method

detlefreichl:


gtk3/lib/gtk3/container.rb
```diff
@@ -16,6 +16,7 @@
 
 module Gtk
   class Container
+    alias_method :<<, :add
```
kou: Please define a new method instead of `alias_method` such as:

```ruby
def <<(widget)
  add(widget)
end
```

The code uses raw `add` method. `<<` is a convenience method. So it's not natural that `<<` uses raw `add` method.

PR #290, opened 2015-01-04
add <<-alias for add-method

detlefreichl:
changed, that it uses the add-method of Gtk::Container


kou:
Thanks! I've merged it!

PR #292, opened 2015-01-10
add Gtk::Container foreach and forall methods

detlefreichl:
The foreach and forall methods do not have the same names like the ones in the gtk-no-gi branch (each and each_forall), but they are direct mappings to the original gtk+ methods, which I think is in the spirit of the gi bindings. Maybe additional aliases should be created.

### DIFF
--- a/gtk3/ext/gtk3/depend
+++ b/gtk3/ext/gtk3/depend
@@ -1,0 +1,6 @@
+install: install-pc
+install-pc:
+	if test -n "$(pkgconfigdir)"; then			\
+	  $(MAKEDIRS) $(pkgconfigdir);				\
+	  $(INSTALL_DATA) ruby-gtk3.pc $(pkgconfigdir);		\
+	fi
--- a/gtk3/ext/gtk3/extconf.rb
+++ b/gtk3/ext/gtk3/extconf.rb
@@ -1,0 +1,77 @@
+#!/usr/bin/env ruby
+#
+# Copyright (C) 2015  Ruby-GNOME2 Project Team
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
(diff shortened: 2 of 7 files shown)


kou:
Thanks for your pull request.
Your code is almost good but Ruby/GObjectIntrospection has API to register just callback. So we don't need to bind each functions.

I implement it. Now, you can use `Container#each` and `Container#each_all` at master.

detlefreichl:
Am 11.01.2015 um 03:24 schrieb Kouhei Sutou:
> Thanks for your pull request.
> Your code is almost good but Ruby/GObjectIntrospection has API to
> register just callback. So we don't need to bind each functions.

I implemented it after

gio2/ext/gio2/rb-gio2-pollable-source.c

Is it possible to change this one to gi-callback, too?


> 
> I implement it. Now, you can use |Container#each| and
> |Container#each_all| at master.
> 
Thank you.

Because the methods do almost the same, what do you think about the
following API? Maybe as an alternative implementation.

Container#each(all=false)

"each" called with no args or "false" will call the C _foreach function
and "each" called with "true"-argument will call the C _forall function.

kou:
> gio2/ext/gio2/rb-gio2-pollable-source.c
>
> Is it possible to change this one to gi-callback, too?

No. We can't use the approach to rb-gio2-pollable-source.c. Because it just sets callback proc. instead of calling a function with callback.



> Because the methods do almost the same, what do you think about the following API? Maybe as an alternative implementation.
>
> Container#each(all=false)
>
> "each" called with no args or "false" will call the C _foreach function and "each" called with "true"-argument will call the C _forall function.

Thanks for your suggestion.
But it's not Rubyisth convension, I think. Because built-in Ruby classes provides `each` family methods for different `each` behavior instead of passing parameter to one `each` method. For example, `IO` provides `each`, `each_line`, `each_char`, `each_byte` and so on.

PR #305, opened 2015-02-21
update aboutdialog.rb

cedlemo:
First update, I haven't modified / updated the  comments that are related to the copyright. This script complains that this kind of windows should have a parent window but I didn't add one in order to stay close to the original example.


kou:
Thanks!
I've merged and improved the sample.

Could you update copyright year at the next time?

And could you also try to fix the following?

> This script complains that this kind of windows should have a parent window

PR #309, opened 2015-02-23
update copyright and version check

cedlemo:
I have just updated the copyright and the version check because it looks like that the show Class method doesn't exist. I will create an issue for this


kou:
Thanks. I've merged it.

Please remove existing copyright notice at the next time. It's required information.

cedlemo:
Sorry Kou, but I don't understand why I should remove copyright notice if it's a required information.

kou:
Ah! Sorry! "You should *NOT* remove existing copyright notice"...

PR #315, opened 2015-02-26
finish aboutdialog2.rb, update copyright version check for textbuffer_serialize.rb

cedlemo:
For aboutdialog2.rb I had have to remove the parts related to the email/url hook. Those methods were deprecated class methods. In order to update those parts I would have to use signal_connect which is an instance method so I would have to complicated the example.

### DIFF
--- a/gtk3/sample/misc/aboutdialog2.rb
+++ b/gtk3/sample/misc/aboutdialog2.rb
@@ -12,15 +12,6 @@ unless Gtk::Version.or_later?(3, 4, 2)
   exit
 end
 
-Gtk::AboutDialog.set_email_hook {|about, link|
-  p "email_hook"
-  p link
-}
-Gtk::AboutDialog.set_url_hook {|about, link|
-  p "url_hook"
-  p link
-}
-
 Gtk::AboutDialog.show(nil,
 		      "artists" => ["Artist 1 <no1@foo.bar.com>", "Artist 2 <no2@foo.bar.com>"],
 		      "authors" => ["Author 1 <no1@foo.bar.com>", "Author 2 <no2@foo.bar.com>"],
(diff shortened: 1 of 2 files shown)


gtk3/sample/misc/aboutdialog2.rb
```diff
@@ -12,15 +12,6 @@
   exit
 end
 
-Gtk::AboutDialog.set_email_hook {|about, link|
```
kou: Could you use `active-link` signal instead of just removing these methods?

gtk3/sample/misc/textbuffer_serialize.rb
```diff
@@ -2,21 +2,18 @@
 =begin
   textbuffer_serialize.rb - Ruby/GTK sample script.
 
-  Copyright (c) 2006 Ruby-GNOME2 Project Team 
+  Copyright (c) 2005-2015 Ruby-GNOME2 Project Team
```
kou: Please don't change copyright start year.

cedlemo: Oups sorry. You can reject this merge request . I will make the modifications after that

gtk3/sample/misc/aboutdialog2.rb
cedlemo: I don't know because I have to create an instance in order to add a signal and the show method is a class method so I think that the dialog that is shown and the dialog that is connected to the signal won't be the same. Maybe I should try to use something like mydialog.class.show.

kou: I understand. Please ignore this request.

gtk3/sample/misc/textbuffer_serialize.rb
kou: OK. I closed this pull request.

PR #339, opened 2015-03-17
update copyright in helloworld.rb

cedlemo:


kou:
Thanks for your works!

PR #341, opened 2015-03-19
update to gtk3 api  and update copyright in checkbutton.rb

cedlemo:


cosmo0920:
Thanks for your works, @cedlemo! It is very helpful for us, too. XD

PR #342, opened 2015-03-19
update to gtk3 api  and update copyright in colorselection.rb

cedlemo:


cosmo0920:
Thank you for your works, @cedlemo! It is very helpful for us. :)

PR #344, opened 2015-03-21
gtk3: use 1-12 as a range of calendar month

myokoym:
GTK+ 2 and GTK+ 3 are use 0-11 as a range of calendar month.
But Ruby/GTK2 is uses 1-12 as a range of calendar month. I
think we should use 1-12 as a range of calendar month.

What do you think?

See: https://github.com/ruby-gnome2/ruby-gnome2/blob/master/gtk2/ext/gtk2/rbgtkcalendar.c#L41

### DIFF
--- a/gtk3/lib/gtk3/calendar.rb
+++ b/gtk3/lib/gtk3/calendar.rb
@@ -1,0 +1,24 @@
+# Copyright (C) 2015  Ruby-GNOME2 Project Team
+#
+# This library is free software; you can redistribute it and/or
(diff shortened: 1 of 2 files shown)


myokoym:
For your information, Ruby's standard library uses 1-12 as a range of month:
 http://ruby-doc.org/core-2.2.1/Time.html#method-i-month

I'll merge it tommorow if there are especially no comments.

kou:
It's reasonable.

Issue #340, opened 2015-03-19
ComboBox error at initialization


cedlemo:
With this code:

```ruby
#!/usr/bin/env ruby
require "gtk3"

#
# Icon and text
#
window = Gtk::Window.new
model = Gtk::ListStore.new(Gdk::Pixbuf, String)
[[Gtk::Stock::QUIT, "quit"],
 [Gtk::Stock::CANCEL, "cancel"],
 [Gtk::Stock::OK, "ok"]].each do |stock, name|
  iter = model.append
  iter[0] = window.render_icon_pixbuf(stock, Gtk::IconSize::MENU)
  iter[1] = name
end

combo2 = Gtk::ComboBox.new(:model => model)

```

I have this error:

```
/home/cedlemo/.gem/ruby/2.2.0/gems/gobject-introspection-2.2.5/lib/gobject-introspection/loader.rb:284:in `invoke': invalid argument Hash (expect #<Class:0x000000030e2330>) (ArgumentError)
	from /home/cedlemo/.gem/ruby/2.2.0/gems/gobject-introspection-2.2.5/lib/gobject-introspection/loader.rb:284:in `block (2 levels) in load_constructor_infos'
	from /home/cedlemo/.gem/ruby/2.2.0/gems/gobject-introspection-2.2.5/lib/gobject-introspection/loader.rb:300:in `block in load_constructor_infos'
	from /home/cedlemo/.gem/ruby/2.2.0/gems/glib2-2.2.5/lib/glib2/deprecatable.rb:47:in `block in define_deprecated_method_by_hash_args'
	from combobox.rb:17:in `new'
	from combobox.rb:17:in `<main>'
```

cedlemo:
I think it is the same kind of issue that #320 I would like try to solve it if you have no objections.

kou:
Sorry for my late response. :<
Good work. :-)

PR #345, opened 2015-03-21
add hash arguments for Gtk::ComboBox.new

cedlemo:
This is for the issue #340 
I think that the combo-box.rb could be modified/improved. I the entry parameter(line 21 of this file) is  a boolean false by default.

I am not an experienced dev sorry. I would appreciate any advices.


kou:
Thanks for the work!
It's almost good. :-)

I've merged it with some modifications.

PR #346, opened 2015-03-22
handle initialize with entry for Gtk::ComboBoxText

cedlemo:
Allow user to create a Gtk::ComboBoxText with entry using :entry => true

### DIFF
--- a/gtk3/test/test_gtk_combo_box_text.rb
+++ b/gtk3/test/test_gtk_combo_box_text.rb
@@ -1,0 +1,38 @@
+# Copyright (C) 2015  Ruby-GNOME2 Project Team
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


gtk3/test/test_gtk_combo_box_text.rb
```diff
@@ -0,0 +1,38 @@
...
+      test "no others" do
+        combo_box_text = Gtk::ComboBoxText.new(:entry => true)
+        assert do
+          combo_box_text.has_entry?
+        end
+      end
+    end
+    sub_test_case("no :entry") do
+      test "no others" do
+        combo_box_text = Gtk::ComboBox.new
```
kou: Typo: `Text` is missing in `Gtk::ComboBox`.

cedlemo:
Oups sorry !

I was wondering if I should remove the<code>sub_test_case</code> for  :entry true or false and just create two <code> test "entry =>true"</code> and <code>test "entry => false"</code>.

kou:
> I was wondering if I should remove the `sub_test_case` for `:entry true` or `false` and just create two `test "entry =>true"` and `test "entry => false"`.

It's better. Could you do so?

cedlemo:
Done  .

kou:
Merged!

PR #347, opened 2015-03-22
update to gtk3 api  and update copyright in combobox.rb

cedlemo:


kou:
Thanks!
I've merged and added `#set_attributes`.

PR #354, opened 2015-03-25
Gtk::ColorSelectionDialog with hash arguments

cedlemo:
I had had to add an empty string if no title was provided otherwise the test complains with:

    TypeError: no implicit conversion of nil into String

### DIFF
--- a/gtk3/lib/gtk3/color-selection-dialog.rb
+++ b/gtk3/lib/gtk3/color-selection-dialog.rb
@@ -1,0 +1,26 @@
+# Copyright (C) 2015  Ruby-GNOME2 Project Team
+#
+# This library is free software; you can redistribute it and/or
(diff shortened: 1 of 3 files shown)


kou:
Thanks for your work!
But we don't support `Gtk::ColorSelectionDialog` because [gtk_color_selection_dialog_new()](https://developer.gnome.org/gtk3/stable/GtkColorSelectionDialog.html#gtk-color-selection-dialog-new) is deprecated in GTK+ 3.

Sorry...

kou:
FYI: GTK+ introduces `Gtk::ColorChooserDialog` for color choosing widget.

cedlemo:
Yes I was aware of ColorChooserDialog. It is just that there were no deprecated error messages and I didn't search further. It is  me who is sorry. Next time I will check.

kou:
Thanks!

PR #355, opened 2015-03-26
Gtk::FontChooserDialog with hash arguments

cedlemo:
In order to be consistant with the other dialogs


kou:
Thanks! Sorry for my late response.

PR #356, opened 2015-03-26
add deprecated messages for Gtk::ColorSelection 

cedlemo:
I hope this is the right way to do this

### DIFF
--- a/gtk3/lib/gtk3/deprecated.rb
+++ b/gtk3/lib/gtk3/deprecated.rb
@@ -6,6 +6,7 @@ module Gtk
   define_deprecated_const :FileSelection,          :raise => "Use 'Gtk::FileChooserDialog' instead."
   define_deprecated_const :FontSelection,          :raise => "Use 'Gtk::FontChooserWidget' instead."
   define_deprecated_const :FontSelectionDialog,    :raise => "Use 'Gtk::FontChooserDialog' instead."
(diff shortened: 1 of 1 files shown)


gtk3/lib/gtk3/deprecated.rb
```diff
@@ -6,6 +6,8 @@ module Gtk
   define_deprecated_const :FileSelection,          :raise => "Use 'Gtk::FileChooserDialog' instead."
   define_deprecated_const :FontSelection,          :raise => "Use 'Gtk::FontChooserWidget' instead."
   define_deprecated_const :FontSelectionDialog,    :raise => "Use 'Gtk::FontChooserDialog' instead."
+  define_deprecated_const :ColorSelection          :raise => "Use 'Gtk::ColorChooserWidget' instead."
```
kou: Add `remove_const(:ColorSelection)` before the line. `define_deprecated_const` uses `const_missing` hook. So we can't have defined constant.

cedlemo:
Ok just to be sure that I understand why I am  about to do:

As the ColorSelection exist in the module, we need first to use the remove_const in order to remove all definition of this class so when we call Gtk::ColorSelection this trigger the const_missing which use the define_deprecated_const :ColorSelection (Am I right?)

kou:
Right!

kou:
Thanks.

PR #359, opened 2015-03-31
Gtk3 sample misc label rb

cedlemo:


cedlemo:
Bad pull request sorry

PR #362, opened 2015-04-03
update to gtk3 api  and update copyright in textbuffer_serialize.rb

cedlemo:
There are a lot of changes between the previous api and the current so I had to modify somehow the script. I think that it is close to the previous one. Let me know what you want me to rewritte.

### DIFF
--- a/gtk3/sample/misc/textbuffer_serialize.rb
+++ b/gtk3/sample/misc/textbuffer_serialize.rb
@@ -20,7 +20,6 @@ textview = Gtk::TextView.new
 textview.set_size_request(600, 400)
 buffer = textview.buffer
 buffer.text = DATA.read
-serialize_format = buffer.register_serialize_tagset(nil)
 deserialize_format = buffer.register_deserialize_tagset(nil)
 format = buffer.serialize_formats[0]
 
--- a/gtk3/sample/misc/textbuffer_serialize.rb
+++ b/gtk3/sample/misc/textbuffer_serialize.rb
@@ -20,100 +20,107 @@ textview = Gtk::TextView.new
 textview.set_size_request(600, 400)
 buffer = textview.buffer
 buffer.text = DATA.read
+serialize_format = buffer.register_serialize_tagset(nil)
+deserialize_format = buffer.register_deserialize_tagset(nil)
 format = buffer.serialize_formats[0]
-buffer.register_serialize_tagset(nil)
-buffer.register_deserialize_tagset(nil)
 
 window = Gtk::Window.new("Gtk::TextBuffer Serialize Demo")
 vbox = Gtk::Box.new(:vertical, 0)
 
-serialize_button = Gtk::FileChooserButton.new("Serialize to a file", 
-                                              Gtk::FileChooserAction::OPEN)
-deserialize_button = Gtk::FileChooserButton.new("Serialize to a file", 
-                                                Gtk::FileChooserAction::OPEN)
-
-serialize_button.current_folder = current_folder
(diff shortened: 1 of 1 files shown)


gtk3/sample/misc/textbuffer_serialize.rb
```diff
@@ -20,100 +20,107 @@
 textview.set_size_request(600, 400)
 buffer = textview.buffer
 buffer.text = DATA.read
+serialize_format = buffer.register_serialize_tagset(nil)
```
kou: Is this line needed? It seems that `serialize_format` isn't used.

gtk3/sample/misc/textbuffer_serialize.rb
```diff
@@ -20,100 +20,106 @@
...
-  dialog = Gtk::FileChooserDialog.new("Deserialize from a file",
-                                      window,
-                                      Gtk::FileChooserAction::OPEN,
-                                      nil,
+toolbar.set_style(Gtk::ToolbarStyle::BOTH)
+button_open = Gtk::ToolButton.new(:icon_widget => nil, :label => "Deserialize from a file",:stock_id => Gtk::Stock::OPEN)
+button_open.signal_connect "clicked" do
+  dialog = Gtk::FileChooserDialog.new(:title => "Deserialize from a file",
+                                      :parent => window,
+                                      :actions => Gtk::FileChooserAction::OPEN, #or :open
```
kou: `:open` is better!

gtk3/sample/misc/textbuffer_serialize.rb
```diff
@@ -20,100 +20,106 @@
...
+toolbar.insert(button_clear, 0)
+
+button_color = Gtk::ToolButton.new(:icon_widget => nil, :label => "Color the region",:stock_id => Gtk::Stock::SELECT_COLOR)
+button_color.signal_connect "clicked" do 
+  dialog = Gtk::ColorChooserDialog.new(:title => "Color the region", :parent => window)
   if dialog.run == Gtk::ResponseType::OK
-    bounds = buffer.selection_bounds
-    color = dialog.colorsel.current_color
-    tag_name = color.to_a.inspect
+    bounds = buffer.selection_bounds # returns an array [test_selected(t/f)? ,Gtk::TextIter, Gtk::TextIter]
```
kou: I improved the API at master. This returns `[Gtk::TextIter, Gtk::TextIter]` at master.

gtk3/sample/misc/textbuffer_serialize.rb
cedlemo: does it works with others OPEN constants ? (like Gtk::Stock::OPEN for example)

kou: Yes. To be more precise, nick name of GEnum or GFlags by Symbol is better.
(`Gtk::Stock::OPEN` isn't good example because it's just a macro not GEnum.)

For example, both `Gtk::Window.new(Gtk::WindowType::POPUP)` and `Gtk::Window.new(:popup)` work. We want to recommend the latter. Because users don't need to remember enum namespace (`Gtk::WindowType` in this example).

cedlemo:
I have added more commits (see here https://github.com/ruby-gnome2/ruby-gnome2/compare/master...cedlemo:gtk3_sample_misc_text_buffer_serialize) but they aren't displayed it seems.

kou:
It's strange...
I'll try to merge this pull request. If some commits aren't merged, could you send them as another pull request?

kou:
It seems that some commits aren't merged...

PR #364, opened 2015-04-04
update to gtk3 api  and update copyright in iconview.rb

cedlemo:


kou:
Thanks!
I've added support of `Gtk::IconView.new(model)` to master.

PR #371, opened 2015-04-11
update to new api and update copyright linkbutton.rb

cedlemo:
I had to remove the part Gtk::LinkButton.set_uri_hook because it was deprecated.


kou:
Thanks!

PR #377, opened 2015-04-19
update to new api and update copyright tooltips.rb

cedlemo:


gtk3/sample/misc/tooltips.rb
```diff
@@ -5,28 +5,27 @@
...
+  puts "This sample requires GTK+ 3.4.2 or later: #{Gtk::Version::STRING}"
+  exit
 end
 
+
 def treeview_query_tooltip(treeview, keyboard_tip, x, y, tooltip)
     if keyboard_tip
         # Keyboard mode
-        path, = treeview.cursor
+        ok,path = treeview.cursor
```
kou: It seems that `path, column = treeview.cursor` is right because `gtk_tree_view_get_cursor()` returns `GtkTreePath *` and `GtkTreeViewColumn *`.

kou: FYI: https://developer.gnome.org/gtk3/stable/GtkTreeView.html#gtk-tree-view-get-cursor

gtk3/sample/misc/tooltips.rb
```diff
@@ -5,28 +5,27 @@
...
         if not path
             return false
         end
     else
         bin_x, bin_y = treeview.convert_widget_to_bin_window_coords(x, y)
         # Mouse mode
-        path, = treeview.get_path_at_pos(bin_x, bin_y)
+        ok, path = treeview.get_path_at_pos(bin_x, bin_y)
```
kou: It seems that `get_path_at_pos` returns `[path, column, cell_x, cell_y]`.

FYI: https://developer.gnome.org/gtk3/stable/GtkTreeView.html#gtk-tree-view-get-path-at-pos

cedlemo: Yes for the C api but  I changed because if I let the initial line, I had an error in line 34 saying invalid argument TrueClass (expect Gtk::TreePath). The path variable was a boolean. If I print it I have an array with : [true, #<Gtk::TreePath:0x2b3dd30 ptr=0x3c7ef20 own=true>, #<Gtk::TreeViewColumn:0x2e4d1d8 ptr=0x3d27d90>, 319, 0]

gtk3/sample/misc/tooltips.rb
cedlemo: For this one you are right, I was misled by the other error  below which stop the script when I tryed to update it

gtk3/sample/misc/tooltips.rb
kou: It should be fixed in library. I fixed it. Now `#get_path_at_pos` returns `[Gtk::TreePath, Gtk::TreeViewColumn, cell_x, cell_y]` if path is found, `nil` otherwise.

Could you update the code?

cedlemo:
rebase done.

kou:
Thanks!

PR #379, opened 2015-04-20
use hash argument for Gtk::ToggleButton#new

cedlemo:
I didn't find a way to test the use of mnemonic in the test file. Any ideas?

### DIFF
--- a/gtk3/lib/gtk3/toggle-button.rb
+++ b/gtk3/lib/gtk3/toggle-button.rb
@@ -19,14 +19,12 @@ module Gtk
     alias_method :initialize_raw, :initialize
     def initialize(options={})
       label    = options[:label]
-      mnemonic = options[:mnemonic]
-      if options.empty?
-        initialize_raw
-      elsif label
-        mnemonic ? initialize_new_with_mnemonic(label) :
+      use_underline = options[:use_underline]
+      if label
+        use_underline ? initialize_new_with_mnemonic(label) :
                    initialize_new_with_label(label)
       else
-        raise ArgumentError, "Invalid arguments."
+        initialize_raw
       end
     end
   end
--- a/gtk3/lib/gtk3/toggle-button.rb
+++ b/gtk3/lib/gtk3/toggle-button.rb
@@ -1,0 +1,33 @@
+# Copyright (C) 2015  Ruby-GNOME2 Project Team
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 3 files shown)


gtk3/lib/gtk3/toggle-button.rb
```diff
@@ -0,0 +1,33 @@
...
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+module Gtk
+  class ToggleButton
+    alias_method :initialize_raw, :initialize
+    def initialize(options={})
+      label    = options[:label]
+      mnemonic = options[:mnemonic]
```
kou: Please use `:use_underline` instead of `:mnemonic` because other class such as `Gtk::Label` uses `:use_underline`.

gtk3/lib/gtk3/toggle-button.rb
```diff
@@ -0,0 +1,33 @@
...
+      if options.empty?
```
kou: How about the following code? It's more simpler.

```ruby
if label
  # ... initialize_new_with_label or mnemonic
else
  initialize_raw
end
```

kou:
> I didn't find a way to test the use of mnemonic in the test file. Any ideas?

You can use `use-underline` property. Try `button.use_underline?`.

gtk3/lib/gtk3/toggle-button.rb
cedlemo: At first it is what I wanted to do but when I tryed it, the label was not underlined. For example the label "_test" was displayed as "test" without underline. And on here https://developer.gnome.org/gtk3/stable/GtkToggleButton.html#gtk-toggle-button-new-with-mnemonic you can see that they just say that this function create a mnemonic but not that it underlines the label.

kou: In my understanding, `:use_underline` means that "I use underline in label to indicate mnemonic" not "I expect that label is displayed with underline".

cedlemo: I checked the pyGtk  (https://developer.gnome.org/pygtk/unstable/class-gtktogglebutton) and you are right.

kou: Thanks for confirming it.
Could you update your code?

kou:
Thanks for updating your code!

PR #394, opened 2015-05-03
update to gtk3 api  and update copyright in togglebutton.rb

cedlemo:
It was because of this sample that I was reluctant to use the symbol :use_underline and that I used :mnemonic. On my system when I launch it, the button with :use_underline => true is not underlined but the mnemonic works.

regards


kou:
Thanks!

PR #398, opened 2015-05-04
Gtk3 sample misc cursor rb

cedlemo:


kou:
Thanks.
I changed `Gdk::Cursor.new`. We can omit `Gdk::Display` when we use `Gdk::Display.default`.

PR #402, opened 2015-05-08
add a new sample illustrating the use of Gtk::Application

cedlemo:
Gtk::Application and Gtk::ApplicationWindow example.

### DIFF
--- a/gtk3/sample/misc/application.rb
+++ b/gtk3/sample/misc/application.rb
@@ -1,0 +1,31 @@
+#!/usr/bin/env ruby
+=begin
+  application.rb - Ruby/GTK version of the example-1.c at
+  https://developer.gnome.org/gtk3/stable/gtk-getting-started.html
+  Copyright (c) 2015  Ruby-GNOME2 Project Team
+  This program is licenced under the same licence as Ruby-GNOME2.
+=end
+
+require "gtk3"
+myapp = Gtk::Application.new("org.gtk.example", 0)
+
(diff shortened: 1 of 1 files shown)


kou:
Good work! It's needed for GTK+ 3 support!

cedlemo:
What do you mean by "It's needed for GTK+ 3 support!". Is there anything else that you need for GTK+3 support (I can add them on my list of things to do).

kou:
> What do you mean by "It's needed for GTK+ 3 support!".

Sorry for my lack of explanation. Yes, we're providing many GTK+ 3 API but I think that users also need a good example to use new GTK+ 3 API. Your work can be used for it.

> Is there anything else that you need for GTK+3 support (I can add them on my list of things to do).

Thanks for your help!
Short answer: There is no more tasks except we're working on but ...

Long answer:

I think that the followings are needed for minimal GTK+ 3 support:

  * GTK+ 3 API implementation
  * GTK+ 3 API examples

I'm working on "GTK+ 3 API implementation" with your help.

You're working on "GTK+ 3 API examples".

So we'll release a new version when you migrate all scripts under `sample/`. (API implementation will be done when you migrate all scripts.)

I think that the followings are needed for full GTK+ 3support:

  * API reference
  * Tutorial

They are useful for users but they are additional tasks for us. So we'll release a new version without them.

cedlemo:
Ok it is clear for me now. thx

PR #408, opened 2015-05-10
re-enable use of a symbol as dialogs response-id

detlefreichl:


kou:
Could you show an use case?

detlefreichl:
I had not thought it to the end, so I will close this request for now.

PR #409, opened 2015-05-10
add Gtk::Clipboard#set_text and Gtk::Clipboard#text=

detlefreichl:


gtk3/lib/gtk3/clipboard.rb
```diff
@@ -0,0 +1,25 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+module Gtk
+  class Clipboard
+    alias_method :set_text_raw, :set_text
+    def set_text(text)
+      set_text_raw(text, text.length)
```
kou: Use `text.bytesize` instead. Because `text.length` is the number of characters. GTK+ requires the number of bytes.

detlefreichl:
Changed to bytesize method usage.

kou:
Thanks for changing it.

PR #410, opened 2015-05-10
add hash parameters to Gtk::Paned#pack[12]

detlefreichl:


gtk3/lib/gtk3/paned.rb
```diff
@@ -0,0 +1,33 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+module Gtk
+  class Paned
+    alias_method :pack1_raw, :pack1
+    def pack1(child, options={})
+      resize = options[:resize] || true
```
kou: `pack1(child, :resize => false)` is treated as `pack1(child, :resize => true)`.

detlefreichl:
Now with fix, that hash parameters are evaluated the right way.

kou:
Thanks.

I changed the default value of `:resize` of `#pack1` to `false` from `true` because `gtk_paned_add1(widget)` is equivant to `gtk_paned_pack1(widget, FALSE, TRUE)`.

See also: https://developer.gnome.org/gtk3/stable/GtkPaned.html#gtk-paned-add1

PR #419, opened 2015-05-15
Update to Gtk3 api and update copyright in misc_button.rb

cedlemo:
I have rewrite this initial line : Gtk::Window.new.add(box).set_default_size(200, 100).show_all in multiple lines for a better readability even if I have seen that now this kind of statement works.


kou:
> I have rewrite this initial line : Gtk::Window.new.add(box).set_default_size(200, 100).show_all in multiple lines for a better readability even if I have seen that now this kind of statement works.

Good work!
I also like multiple lines style.

PR #422, opened 2015-05-17
Remove scripts for deprecated UIManager

cedlemo:
The UIManager, ActionGroup are deprecated. 
*    https://developer.gnome.org/gtk3/stable/GtkUIManager.html
*    https://developer.gnome.org/gtk3/stable/GtkActionGroup.html#gtk-action-group-add-radio-actions

Gtk::Builder should be used instead. I will provide an example based on 

https://git.gnome.org/browse/gtk+/tree/examples/builder.c?h=gtk-3-16


kou:
> Gtk::Builder should be used instead. I will provide an example based on
> 
> https://git.gnome.org/browse/gtk+/tree/examples/builder.c?h=gtk-3-16

Thanks!

PR #423, opened 2015-05-17
Add Gtk::Builder script example with its UI file

cedlemo:


cedlemo:
I didn't look to this before create the script, but it seems that the gtk_builder_new_from_{file,ressource,string) exist. Do you want that I add an initialize method in the gtk3/lib/gtk3/builder.rb with hash arguments ?

kou:
> Do you want that I add an initialize method in the gtk3/lib/gtk3/builder.rb with hash arguments ?

Yes!

PR #426, opened 2015-05-21
Gtk3 builder initialize

cedlemo:
As you can see I have added some files and a directory for the tests. I didn't know how it should be done so I wait for advices from you in order to improve this if needed.

If you want I can add a second builder example in the sample script that is based on the use of a resource file. (see here http://stackoverflow.com/questions/30308748/what-is-a-resource-path-that-have-to-be-used-with-gtk-builder-new-from-resource )


kou:
Thanks!

We're using `test/fixture/` for the propose instead of `test/data/`. I've merged and changed it.

> If you want I can add a second builder example in the sample script that is based on the use of a resource file. (see here http://stackoverflow.com/questions/30308748/what-is-a-resource-path-that-have-to-be-used-with-gtk-builder-new-from-resource )

Please do it! It'll be helpful for users.

PR #436, opened 2015-06-04
Add new example for Gtk::FlowBox

cedlemo:


kou:
Great!
Could you add copyright notice of the original source?

cedlemo:
I am sorry I don't see what you means I know that their License can be found here(https://github.com/sebp/PyGObject-Tutorial/blob/master/COPYING). But I don't really know how to write/format it as a copyright reference.

By the way they have done a really good job there (https://github.com/sebp/PyGObject-Tutorial).

kou:
Thanks for showing copyright information of the original source.

I've added copyright notice of the original source.
The original source is licensed under GFDL 1.3 or later. It's not compatible Ruby-GNOME2. Ruby-GNOME2 is licensed under LGPL 2.1 or later. So we should use GFDL 1.3 or later for the script.

PR #440, opened 2015-06-05
Update to Gtk3 api, update Copyright and fix style issue in print.rb

cedlemo:


gtk3/sample/misc/print.rb
```diff
@@ -1,245 +1,246 @@
...
-    end
+  def height_in_points
+    size[1] / Pango::SCALE
+  end
 
-    def width_in_points=(width)
-        self.width = width * Pango::SCALE
-    end
+  def width_in_points=(width)
+    width *= Pango::SCALE
```
kou: It enbugs. `self.width =` is needed. The code just assigns to the local variable.

cedlemo:
Error fixed and rebase done

kou:
Thanks!

PR #449, opened 2015-06-11
Update to Gtk3 api, update Copyright and fix style issue in scalebutt…

cedlemo:
…on.rb

### DIFF
--- a/gtk3/sample/misc/scalebutton.rb
+++ b/gtk3/sample/misc/scalebutton.rb
@@ -10,7 +10,7 @@ require "gtk3"
 
 window = Gtk::Window.new
 window.set_size_request(300, 300)
-scale = Gtk::ScaleButton.new(Gtk::IconSize::BUTTON, 0, 100, 2, nil)
+scale = Gtk::ScaleButton.new
 scale.set_value(30)
 scale.set_valign(Gtk::Align::END)
 scale.set_halign(Gtk::Align::END)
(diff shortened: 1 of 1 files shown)


gtk3/sample/misc/scalebutton.rb
```diff
@@ -2,18 +2,30 @@
...
 
-require 'gtk3'
+require "gtk3"
 
 window = Gtk::Window.new
-window.add(scale = Gtk::ScaleButton.new(Gtk::IconSize::IconSize::BUTTON))
-scale.set_icons(['gtk-goto-bottom', 'gtk-goto-top', 'gtk-execute'])
-scale.signal_connect('value-changed') { |widget, value| puts "value changed: #{value}" }
+window.set_size_request(300, 300)
+scale = Gtk::ScaleButton.new(Gtk::IconSize::BUTTON, 0, 100, 2, nil)
```
kou: It seems that all arguments should be omitable. And they should be able to be specified by `Hash` such as `:icon_size => :button`.

I'll do it tomorrow. If you can do it before I do it, could you create a pull request?

cedlemo:
> it seems that all arguments should be omitable. And they should be able to be specified by Hash

No they weren't when I updated the script. I have created a pull request ( #450 ) for this. 

> :icon_size => :button

I can't use :button , only Gtk::IconSize::BUTTON.

kou:
Thanks! I've merged!

PR #450, opened 2015-06-12
Gtk3 scalebutton initialize

cedlemo:


kou:
Thanks!
I've added `:icon_size => :menu` style support.

PR #451, opened 2015-06-13
Update to Gtk3 api, update Copyright and fix style issue in dnd.rb

cedlemo:
I have added those lines : 

```ruby
         puts selection_data.data[0].pack("C#{data_len}")
         puts selection_data.text
```
Because that was not obvious. If you prefer that I remove all the lines I have added and that I stay closer to the initial script just tell me and I will do the modifications.


kou:
> I have added those lines :
> ...
> Because that was not obvious.

Good idea!

PR #452, opened 2015-06-13
Gtk3 sample misc recent chooser dialog rb

cedlemo:


gtk3/sample/misc/recentchooserdialog.rb
```diff
@@ -39,7 +37,7 @@
       puts info.uri_display
       puts info.age
       puts info.local?
-      puts info.exist?
+      info.exists
```
kou: Please use `exist?`. And don't remove `puts`.

cedlemo: > And don't remove puts.

Sorry that was an error. I have done a rebase

kou:
Thanks for applying my comments!

PR #455, opened 2015-06-15
Remove list_ from Gtk::Widget#list_style_properties

cedlemo:
I tryed to do like this 8e23852b2c2d7349f85d8b8a70d1a1fff5d36424. I am not sure on how to create tests for this


kou:
Thanks.
I've added a test and improved API.

PR #466, opened 2015-06-26
Setup the model of the returned Gtk::TreeIter

cedlemo:
Setup the model of the resulting `Gtk::TreeIter`  returned by `Gtk::ListStore#prepend`

Fix this kind of pb:

```ruby
model = Gtk::ListStore.new(TrueClass, String)
model.prepend.set_values([true, "item5"])
```

that returns:

```
/home/cedlemo/.gem/ruby/2.2.0/gems/gtk3-2.2.6/lib/gtk3/tree-iter.rb:38:in `set_values': undefined method `set_values' for nil:NilClass (NoMethodError)
```

### DIFF
--- a/gtk3/lib/gtk3/list-store.rb
+++ b/gtk3/lib/gtk3/list-store.rb
@@ -49,5 +49,12 @@ module Gtk
       setup_iter(iter)
       iter
     end
+  
+    alias_method :prepend_raw, :prepend
+    def prepend
+      iter = prepend_raw
+      setup_iter(iter)
+      iter
+    end
   end
 end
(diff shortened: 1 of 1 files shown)


cedlemo:
I think that all the methods that return a Gtk::TreePath need to be modified this way. I am okay to add them.Furthermore I think that the test file should be modified/simplified (remove all assert_nothing_raised as you show me in 2e8f5492bee8289cdec8b3105ce73eb5157e611b ). Tell me if you are ok with this and /or if you  have any advices.

kou:
Thanks!

> I think that all the methods that return a Gtk::TreePath need to be modified this way.

You're right.

> I am okay to add them.

It's very helpful. Please do it!

> Furthermore I think that the test file should be modified/simplified (remove all `assert_nothing_raised` as you show me in 2e8f549 ).

You're right. We should remove them.

>Tell me if you are ok with this and /or if you have any advices.

OK! I have no advice. I think that you understand all of them correctly. :-)

PR #467, opened 2015-06-27
Ensure Gtk::TargetEntry in parameters of Gtk::TreeView#enable_model_d…

cedlemo:
Ensure Gtk::TargetEntry in parameters of Gtk::TreeView#enable_model_drag*


kou:
Thanks!

PR #470, opened 2015-07-01
Add a simple script show the API of the Gtk::ListStore

cedlemo:
I don't know if you could find this usefull but I have written this very simple script in order to see how the Gtk::ListStore methods works so I share.

### DIFF
--- a/gtk3/sample/misc/list_store_usage.rb
+++ b/gtk3/sample/misc/list_store_usage.rb
@@ -22,8 +22,8 @@ end
 def load_list
   model = Gtk::ListStore.new(TrueClass, String)
 
-  (1..4).each do |i|
-    model.append.set_values([false, "index#{i - 1}"])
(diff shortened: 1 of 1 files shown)


gtk3/sample/misc/list_store_usage.rb
```diff
@@ -0,0 +1,96 @@
...
+    print iter[1] + "  "
+  end
+  print "\n"
+  puts ""
+end
+
+def load_list
+  model = Gtk::ListStore.new(TrueClass, String)
+
+  (1..4).each do |i|
```
kou: `4.times` is better. You don't need to do `#{i - 1}` with it.

kou:
I think that it's useful. :-)
I want to merge it after you apply my comment.

cedlemo:
Done!

kou:
Merged!

PR #471, opened 2015-07-01
Add example of combobox

cedlemo:
This is an example of a combobox in which the boolean value of the model are represented by a Gtk::CellRenderToggle

### DIFF
--- a/gtk3/sample/misc/combobox_from_cellrender.rb
+++ b/gtk3/sample/misc/combobox_from_cellrender.rb
@@ -1,0 +1,48 @@
+#!/usr/bin/env ruby
+=begin
+  combobox_from_cellrender.rb - Ruby/GTK sample script
+  Copyright (c) 2002-2015 Ruby-GNOME2 Project Team
+  This program is licenced under the same licence as Ruby-GNOME2.
+=end
+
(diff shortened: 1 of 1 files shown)


gtk3/sample/misc/combobox_from_cellrender.rb
```diff
@@ -0,0 +1,48 @@
...
+window.add combobox
+
+combobox.signal_connect("changed") do |widget|
+  list = widget.model
+  list.each do |model, _path, iter|
+    model.set_value(iter, 0, false)
+  end
+  iter = widget.active_iter
+  puts iter.inspect
+  puts "#{list.get_value(iter, 0)} #{list.get_value(iter, 1)}"
```
kou: `iter[0]` and `iter[1]` are recommended.

gtk3/sample/misc/combobox_from_cellrender.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  list.set_value(iter, 0, true)
```
kou: `iter[0] = true` is recommended.

kou:
Thanks!

PR #472, opened 2015-07-01
Update to Gtk3 api, update Copyright and fix style issue in menu.rb

cedlemo:
I have replaced the `Gtk::OptionMenu.new` by a `Gtk::ComboBox` (based on a `Gtk::RenderCellToggle`) like specified in the deprecated message.


kou:
Thanks!
You're right.

PR #478, opened 2015-07-09
pangorenderer.rb  update

cedlemo:
Update to Gtk3 api, update Copyright and fix style issue, change name of the file.

The `Gdk::PangoRenderer` doesn't exist anymore so I decided to change the name of the file.

I modified the script so that it can handle screen change and window resizing.


kou:
> The Gdk::PangoRenderer doesn't exist anymore so I decided to change the name of the file.

I agree with it.

PR #482, opened 2015-07-11
Setup model in returned iter from Gtk::TreeModelFilter#convert_iter_t…

cedlemo:
Setup model in returned iter from Gtk::TreeModelFilter#convert_iter_to_child_iter,

I guess I have the same thing to do for `gtk_tree_model_filter_convert_child_iter_to_iter`


kou:
> I guess I have the same thing to do for `gtk_tree_model_filter_convert_child_iter_to_iter`

Right.

PR #483, opened 2015-07-11
Update to Gtk3 api, update Copyright and fix style issue in treemodel…

cedlemo:
…filter.rb

Works only if the pull request #482 is merged


kou:
Thanks.
I've changed `Gtk::TreeModelFilter#set_modify_func` argument style to variable length arguments from one `Array`. It's compatible with gtk3-no-gi.

PR #484, opened 2015-07-12
Modify README to README.md

prpr-man:
Change text to markdown.

### DIFF
--- a/glib2/README.md
+++ b/glib2/README.md
@@ -1,40 +1,42 @@
-Ruby/GLib2
-==========
+# Ruby/GLib2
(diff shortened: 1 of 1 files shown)


kou:
Thanks!
If you do related changes for other directories such as [atk/README](https://github.com/ruby-gnome2/ruby-gnome2/blob/master/atk/README), [gdk_pixbuf2/README](https://github.com/ruby-gnome2/ruby-gnome2/blob/master/gdk_pixbuf2/README) and so on, we'll merge them. :-)

PR #487, opened 2015-07-14
Add model in iter from Gtk::TreeModelFilter#convert_child_iter_to_iter

cedlemo:


kou:
Thanks but I don't use your implementation. Because variable name is wrong. Sorry.
I've implemented it with test.

cedlemo:
Yes you are right .

I should be more carefull, what you have done make more sense I should not have set "iter" as name for the result of  the method `#convert_child_iter_to_iter`.

Sorry

kou:
Never mind. If your work has some lacks, I can improve your work. :-)

PR #488, opened 2015-07-18
Remove composited-windows.rb, create eventbox.rb

cedlemo:
It seems that the compositing is an outdated technologie and should not be used anymore:

https://developer.gnome.org/gdk3/stable/gdk3-Windows.html#gdk-window-set-composited

I have added an example of `Gtk:EventBox` without compositing.


kou:
Thanks!
I didn't know that `gdk_window_set_composited()` has been deprecated.

cedlemo:
I think that it because compositing is only for X11 whereas Gtk try to provide an API that should work on both X11 (xlib/xcb), and wayland which compositing is not an option : 

*    https://en.wikipedia.org/wiki/Wayland_(display_server_protocol)#Differences_between_Wayland_and_X

*    https://help.gnome.org/misc/release-notes/3.16/developers.html.en (at the end they say that they have improved the support to wayland)

kou:
Thanks for the information!
It seems that Wayland support is a reasonable reason.

Issue #415, opened 2015-05-11
Gtk3 and threads


michelboaventura:
Hi all,

I'm trying to figure out how I'm suposed to work with ruby-gtk3 and threads. I read on some old posts that I should never call a gtk function from outsite gtk's main thread, because it could break things up. Is this still an issue? Also, I've seen [here](https://github.com/ruby-gnome2/ruby-gnome2/blob/2288c76eb2bd9dd8d787d207daa4a122f0983ba8/gtk3-no-gi/sample/misc/threads.rb) an example with two threads outside gtk's main one on which no special care is taken to deal with this.

glurp:
see: http://ruby-gnome2.osdn.jp/hiki.cgi?tips_threads

cedlemo:
@kou : 
Are those informations still up to date?

*    I have checked the ruby forum link in the hiki, it was written in 2007 - 2010. They talk about Gtk2
*    the last modification of the hiki page was in 2010.

If they are, it could be interesting to do a thread script example based on those informations. the thread.rb example that exist doesn't reflect the complexity of the wiki article (I must say that I haven't totally understood it).

Regards

kou:
> Are those informations still up to date?

Yes.

We should not call GTK+ functions from not main thread. I've updated threads.rb: https://github.com/ruby-gnome2/ruby-gnome2/commit/8412c3a57c6d4856b72c3ae32a18124838131dae

PR #489, opened 2015-07-18
Update to Gtk3 api, update Copyright and fix style issue in tree_prog…

cedlemo:
…ress.rb

It works well I don't know if I should have made some modifications in ot order to handle the threads in a better way(#415).


kou:
We should not use thread for this case because GTK+ functions should not be called from not main thread.
I've changed the script to use `GLib::Timeout` instead of `Thread`.

PR #490, opened 2015-07-19
Update to Gtk3 api and update Copyright xcb_curso.rb

cedlemo:
So this is the last one to update.


kou:
Congratulations!
Thanks for your great works!!!

PR #497, opened 2015-07-27
Vte3 terminal feed child

cedlemo:
Sorry kou but I didn't find a way to create a test. I have tried to use GLib::Tiemout , Vte::Terminal#cursor_position with the "cursor-moved" signal but wihtout success. I am stuck.


kou:
Thanks. But I'm busy until the next week...
I'll check your works the week after next...

kou:
Sorry for my late response.
(I found that you're a vi user. See your commit message. :-)

cedlemo:
Yes that make me laugh at first but a little less when I found that I couldn't remove it! -_-'

PR #510, opened 2015-08-17
Add tutorial directory and few samples

cedlemo:


kou:
Great works!
I've added license headers to these scripts.

PR #515, opened 2015-08-22
Gtk3 sample tutorial new files organization

cedlemo:


kou:
Thanks for your quick work!

PR #518, opened 2015-08-24
Update svg-viewer.rb

shevegen:
Hi!

Two tiny changes, this is the first time I am trying this, no idea whether this can work.

Since I don't want to bypass anything, could you have a look at it and say whether this
is ok kou? (I plan to eventually do more changes lateron, so this is mostly a try-first thing)

I replace two "strings" with two symbols instead. Not a big change, but since Symbols
also work I hope this is acceptable.


kou:
Thanks.
If you do more works, please create pull requests.

PR #534, opened 2015-09-10
Add populating the window part

cedlemo:
@kou I need your help in the part "link the template to a custom class widget".

I do understand the principles of GType and or singleton methods and eigenclass in Ruby. But I miss something (or a lot of things) in order to get the big picture and explain it clearly.

*    why do we need to use type_register (create a new GType)
*    why do we overwritte the init method in the eigenclass
*    when this method will be called after that 
*    why the set_template is called in the  init method (why not in the initialize method)
*    Is it more related to GLib:Object ? where can I found more informations on this.

If you have some links / good resources on GType, GObject, GObjectIntrospection that are related to this, I am interested.

### DIFF
--- a/gtk3/sample/tutorial/README.md
+++ b/gtk3/sample/tutorial/README.md
@@ -312,7 +312,7 @@ In this example we create a subclass of `Gtk::Application` called ExampleApp. In
 *    open : opens files and shows them in a new window
 
 For more informations, see [here](https://wiki.gnome.org/HowDoI/GtkApplication).
-In this case, the signal "*activate*" will be triggered if no arguments are given to the `ExampleApp#run` method. And a default window will be created and will be presented to the user ( [see](https://developer.gnome.org/gtk3/stable/GtkWindow.html#gtk-window-present).
+In this case, the signal "*activate*" will be triggered if no arguments are given to the `ExampleApp#run` method. And a default window will be created and will be presented to the user ( [`Gtk::Widget#present`](https://developer.gnome.org/gtk3/stable/GtkWindow.html#gtk-window-present)).
 
 If file names are given to the `ExampleApp#run` method, then it is the "*open*" signal that is called.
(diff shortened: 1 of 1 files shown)


kou:
> * why do we need to use type_register (create a new GType)

Because inheritance in Ruby doesn't always mean "creating a new GType".

For example, the following code doesn't create a new GType like
`MyWindow`:

```ruby
class MyWindow < Gtk::Window
end
```

Normally, Ruby-GNOME2 users doesn't need to create a new GType. So,
Ruby-GNOME2 doesn't create a new GType by default.

If users want to create a new GType, they need to call `type_register`
explicitly.

> * why do we overwritte the init method in the eigenclass
> * when this method will be called after that

Because we need to custom new class such as setting a template when
creating a new class for object system in GObject. The `init` method
is for the propose.

The `init` method is called when a new class for object system in
GObject is created. New class for object system in GObject is created
when it's needed such as the first instance is created.

New class in Ruby is already created when new class for object system
in GObject is created.

> * why the set_template is called in the init method (why not in the initialize method)

Template information is shared in all instances of the class. If we
set template information in `initialize`, we can't share the
information in all instances.

> * Is it more related to GLib:Object ? where can I found more informations on this.

Uhm...

I understood about template in GTK+ 3 by reading GTK+ reference manual
such as

  * [gtk_widget_class_set_template()](https://developer.gnome.org/gtk3/stable/GtkWidget.html#gtk-widget-class-set-template)
  * [gtk_widget_init_template()](https://developer.gnome.org/gtk3/stable/GtkWidget.html#gtk-widget-init-template)
  * and more `gtk_widget_*_template*()` documents.

But they are incomprehensive...

(I might understand the template system by implementing it in Ruby/GTK3. :-)


I hope that these answers help you...

cedlemo:
>I hope that these answers help you...

Yes, it does. Thanks a lot.

kou:
You're welcome.
I've merged this.

PR #535, opened 2015-09-10
Add Vte#Version.or_later? in vte3

cedlemo:


kou:
Thanks!
But the implementation doesn't work because `Vte.check_version` doesn't exist.
I've implemented it.

PR #536, opened 2015-09-10
Fix issue when signal update-words is triggered

cedlemo:


gtk3/sample/tutorial/exampleapp8/exampleapp.rb
```diff
@@ -157,8 +157,8 @@ def initialize(application)
                    sidebar,
                    "reveal-child",
                    Gio::SettingsBindFlags::DEFAULT)
-    sidebar.signal_connect "notify::reveal-child" do |_sidebar, _gparamspec, an_application|
-      update_words(an_application)
+    sidebar.signal_connect "notify::reveal-child" do |_sidebar, _gparamspec, _application|
```
kou: `_application` is needless.

If you want to get the third argument (it's application specific argument), write the following code:

```ruby
sidebar.signal_connect "notify::reveal-child", self do |_sidebar, _gparamspec, an_application|
  # ...
end
```

Rest `signal_connect` arguments (`self` in this case) are passed to block.

cedlemo:
I have done the modifications and a rebase.

kou:
Good!

PR #542, opened 2015-09-12
Fix issue with path value from tooltips.rb

cedlemo:


gtk3/sample/misc/tooltips.rb
```diff
@@ -25,7 +25,7 @@ def treeview_query_tooltip(treeview, keyboard_tip, x, y, tooltip)
   else
     bin_x, bin_y = treeview.convert_widget_to_bin_window_coords(x, y)
     # Mouse mode
-    path = treeview.get_path_at_pos(bin_x, bin_y)
+    path, = treeview.get_path_at_pos(bin_x, bin_y)
```
cosmo0920: I think that using `_` to clarify "not using anymore this value" is better.

```ruby
path, _ = treeview.get_path_at_pos(bin_x, bin_y)
```

cedlemo:
I agree

kou:
I don't agree.
`x, = xxx` means that "I want to use just the first return value".
`x, _ = xxx` means that "I want to use just the first return value in 2 return values".

"in 2 return values" are needless information. We don't care about how many values are returned.

In fact, `treeview.get_path_at_pos` returns 4 values. "in 2 return values" is wrong information. It'll be confused people who read the code.

`x, *_ = xxx` is better than `x, _ = xxx` in this case. But I think that `x, = xxx` is better than `x, *_ = xxx` because it doesn't have needless information (the number of return values).

cosmo0920:
> x, *_ = xxx is better than x, _ = xxx in this case.

I think so too. 

> But I think that x, = xxx is better than x, *_ = xxx because it doesn't have needless information (the number of return values).

But I don't think so. 

It is important to clarify "this value is no longer needed" here. 

I like notation which is clarifying rather than simplify.
But sometimes conflicts Rubyish way of thinking, though....

kou:
> It is important to clarify "this value is no longer needed" here.

Why do you think that `x, = xxx` doesn't say *this value (rest return values) is no longer needed*? The code says that "the first return value is only used".

> I like notation which is clarifying rather than simplify.

I didn't say about simplicity. I said that "I don't care about the rest return values" (`x, = xxx`) is more important rather than "I receives the rest return values but I don't use them" (`x, *_ = xxx`). Because the former is what the writer intends.

cedlemo:
>I don't agree.

ok :)
@kou Do you want I revert the changes I made?

cedlemo:
I removed "_" and made a rebase.

kou:
Thanks!
I've merged.

cosmo0920:
> I said that "I don't care about the rest return values" (x, = xxx) is more important rather than "I receives the rest return values but I don't use them" (x, *_ = xxx).

IMO, I agree former style to conserve memory. If only concerning about clarifying, I disagree former style. Because former style is NOT clarifying about *"I receives the rest return values but I don't use them"*.

But my feeling sometimes conflicts Rubyish way of thinking.
cedlemo, please don't mind my comments. You should obey kou's comments.

kou:
> IMO, I agree former style to conserve memory. If only concerning about clarifying

I only concern about clarifying.

> I disagree former style. Because former style is NOT clarifying about "I receives the rest return values but I don't use them".

I hope that you understand that "I" do NOT want to clarify about "I receives the rest return values but I don't use them". (I understand that "you" want to clarify about it.)
Because I don't think about it. (I also understand that "you" think about it.) "I don't care about rest return values" is what I think and what I want to clarify. Writing nothing about the result return values (`*_`) clarifies it.

----
FYI:

If you want to say about what you think, you don't need to understand what I think. And it's not bad thing. It's OK to me. I just understand that there is another opinion. (But I may not agree with the opinion.)

If you want to change this project, you need to understand what I think. (Normally, project has some/many active members. But this project has a few active members and I'm the most active member. :<)

PR #547, opened 2015-09-14
Remove dependancy to test-unit-notify gem

cedlemo:
All references have been found with:

    find ./ -type f -name "*.rb"| xargs grep -l "unit/notify"


kou:
Thanks!

Issue #308, opened 2015-02-23
Add XXX::Version.or_later? API


kou:
`Gtk::Version.or_later?` is more Ruby-ish rather than `Gtk.check_version`. We should provide Ruby-ish methods.

See `Gtk::Version.or_later?` in gtk3 for details.

kou:
Could someone work on this?

cedlemo:
You mean something like this #535 but for the other module?

kou:
> You mean something like this #535 but for the other modules?

Yes.

> If yes give me a list of the modules I should focus on and I will do it.

Thanks! Here is a list:

  * atk
  * clutter-gstreamer
  * clutter-gtk
  * gdk3
  * gdk_pixbuf2
  * gio2
  * glib2
  * gobject-introspection
  * gstreamer
  * gtk2
  * gtksourceview3
  * pango
  * poppler
  * rsvg2
  * webkit-gtk

cedlemo:
Done ?

kou:
Wow!!! Very thanks for your great works!!!
I close this.

cedlemo:
You are welcome! I have learnt a lot, that was fun.

PR #552, opened 2015-09-18
Atk version or later method

cedlemo:
I hope it is close to what you want. 

I had to remove the test on `vendor_dir.exist?` otherwise I was not able to run the test on my system (Arch Linux x86_64)

### DIFF
--- a/atk/lib/atk.rb
+++ b/atk/lib/atk.rb
@@ -1,3 +1,19 @@
+# Copyright (C) 2015  Ruby-GNOME2 Project Team
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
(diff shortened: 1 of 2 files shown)


atk/test/test-version.rb
```diff
@@ -0,0 +1,53 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+class TestAtkVersion < Test::Unit::TestCase
+  include AtkTestUtils
+
+  setup do
+    unless Atk::Version.const_defined?(:STRING)
```
kou: The check is needless. Because Ruby/ATK can always provide it.

atk/lib/atk.rb
```diff
@@ -11,11 +27,24 @@
   require 'atk.so'
 end
 
-if vendor_dir.exist?
```
kou: We shouldn't remove the check.
Because atk gem doesn't depend on gobject-introspection gem on non Windows environment. If it's removed, `require "atk"` raises `LoadError on `gem install atk` only environment (no `gem install gobject-introspection` environment).

kou:
> I hope it is close to what you want.

Thanks! It's almost what I think!
I added some comments. Please check them.

> I had to remove the test on `vendor_dir.exist?` otherwise I was not able to run the test on my system (Arch Linux x86_64)

Could you show the error message in the case?
(Oh! You're an Arch Linux user!)

cedlemo:
I haven't any error message, that was the problem because vendor_dir don't exist on my system and I was not able to test it (I forgot that all the gem of ruby-gnome2 runs on other platforms).

cedlemo:
Is it ok? #308

kou:
Thanks!
I've merged it.

> I haven't any error message, that was the problem because vendor_dir don't exist on my system and I was not able to test it (I forgot that all the gem of ruby-gnome2 runs on other platforms).

Thanks for the description. I understoond.

> Is it ok? #308

Yes!

Issue #551, opened 2015-09-18
Latest versions of everything, with MRI 2.2.3p173, cannot install gobject-introspection


dan-higgins:
Hello Ruby friends. I have been trying various things all night, and I simply cannot get this to work. I fear it is a file that isn't in MRI 2.2. Here is some info on my Arch Linux machine:
```
dan@danbook:~/tmp 04:30:19
$ ruby --version
ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-linux]
dan@danbook:~/tmp 04:31:19
$ uname -a
Linux danbook.danamis.com 4.1.6-1-ARCH #1 SMP PREEMPT Mon Aug 17 08:52:28 CEST 2015 x86_64 GNU/Linux
```
Below is from a typescript I made, which has all the details I could think of. The problem seems be a common one, the error:
```
conftest.c:4:18: fatal error: node.h: No such file or directory
```
I hope you can help.

```
Script started on Fri 18 Sep 2015 04:17:06 AM CDT
]0;dan@danbook:~/tmp]0;dan@danbook:~/tmp]2;dan@danbook:~/tmpdan@danbook:~/tmp 04:17:06

$ gem list

*** LOCAL GEMS ***

bigdecimal (1.2.7, 1.2.6)
diff-lcs (1.2.5)
domain_name (0.5.24)
http-cookie (1.0.2)
iconv (1.0.4)
id3lib-ruby (0.6.0)
io-console (0.4.3)
json (1.8.3, 1.8.1)
mechanize (2.7.3)
mime-types (2.6.2)
mini_portile (0.6.2)
minitest (5.8.0)
net-http-digest_auth (1.4)
net-http-persistent (2.9.4)
nokogiri (1.6.6.2)
ntlm-http (0.1.1)
power_assert (0.2.4)
psych (2.0.15, 2.0.8)
rake (10.4.2)
rake-compiler (0.9.5)
rdoc (4.2.0)
rdtool (0.6.38)
rspec (3.3.0)
rspec-core (3.3.2)
rspec-expectations (3.3.1)
rspec-mocks (3.3.2)
rspec-support (3.3.0)
rubygems-update (2.4.8)
test-unit (3.1.3)
unf (0.1.4)
unf_ext (0.0.7.1)
webrobots (0.1.1)
[...]

PR #553, opened 2015-09-18
Add support for native gobject-introspection package installation for…

dan-higgins:
… Arch Linux.

I did confirm that the native system package was installed automatically. I hope it helps.

### DIFF
--- a/glib2/lib/mkmf-gnome2.rb
+++ b/glib2/lib/mkmf-gnome2.rb
@@ -503,6 +503,8 @@ def package_platform
     :redhat
   elsif File.exist?("/etc/SuSE-release")
     :suse
+  elsif find_executable("pacman")
+    :arch
   elsif find_executable("brew")
     :homebrew
   elsif find_executable("port")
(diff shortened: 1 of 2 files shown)


kou:
Thanks!
I've merged it.

Now, gobject-introspecion gem supports auto native package installation on Arch Linux. :-)

There are gems that doesn't support it yet. For example, gtk3 gem doesn't support it:
https://github.com/ruby-gnome2/ruby-gnome2/blob/master/gtk3/ext/gtk3/extconf.rb#L66-L70

We're happy that you send a pull request when you get the same error as #551 with other gem like gtk3 gem. (If you don't get it, you don't need to work on this. But if you work on this, we're happy. ;-))

Thanks for your work for supporting auto GObjectIntrospection package installation on Arch Linux. It helped us.

dan-higgins:
@kou you are welcome. I think the ruby-gnome2 project is lucky to have you. I will work on the gtk3 gem this weekend and submit a pull request for it too. Cheers.

kou:
THanks!

PR #562, opened 2015-09-20
Remove reference to test-unit-notify in some Rakefile

cedlemo:
Sorry but when I checked for the test-unit-notify I just only search them in `*.rb` files

here is the command that returns me the files that remains:

```
find ./ -type f | xargs grep -l "test-unit-notify"
```


kou:
Thanks!

PR #563, opened 2015-09-21
Add Arch Linux support for auto install of native Gtk3 package.

dan-higgins:
Testing this was way harder than changing the code. I had to set up an Arch VM with a bunch of gui packages removed, to get Gtk3 to be removed.

I think it will be rare that users don't already have Gtk3 installed, but at least I have seen this change work.


kou:
Thanks!
Sorry for asking you to work on a fuzzy task...

dan-higgins:
No worries. I'm glad I could help a little. Cheers.

Issue #569, opened 2015-09-24
Gio2 Unable to find version informations


cedlemo:
I can not find any constants or methods in order to get informations on the version of gio

kou:
GIO is included in GLib. So we can version information in GLib:

  * `GLib::MAJOR_VERSION`
  * `GLib::MINOR_VERSION`
  * `GLib::MICRO_VERSION`
  * `GLib::VERSION`

kou:
We just define `GLib::Version` as `Gio::Version`:

```ruby
module Gio
  Version = GLib::Version
end
```

cedlemo:
Thanks

PR #571, opened 2015-09-25
Gio2 version or later

cedlemo:
ref #308 
ref #569

### DIFF
--- a/gio2/lib/gio2.rb
+++ b/gio2/lib/gio2.rb
@@ -52,4 +52,6 @@ module Gio
       loader.load("Gio")
     end
   end
+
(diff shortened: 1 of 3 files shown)


gio2/lib/gio2.rb
```diff
@@ -52,4 +52,6 @@ class << self
       loader.load("Gio")
     end
   end
+
+  Version = GLib::Version
```
cosmo0920: This is just comment. You don't have to take care.
IMO, I want to put this `Version` constant into near another constant definition.

kou:
Thanks!

PR #573, opened 2015-09-25
Gstreamer version or later

cedlemo:
ref #308 

I have added a Gst::Version::NANO constant but I let the Gst::Version::STRING with the same form as in the other module ie: MAJOR . MINOR . MICRO.

### DIFF
--- a/gstreamer/lib/gst.rb
+++ b/gstreamer/lib/gst.rb
@@ -71,6 +71,7 @@ module Gst
       require "gst/registry"
       require "gst/structure"
       require "gst/type-find-factory"
+      require "gst/version"
       init_base
       init_controller
     end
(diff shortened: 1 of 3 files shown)


kou:
Thanks!

> I have added a Gst::Version::NANO constant but I let the Gst::Version::STRING with the same form as in the other module ie: MAJOR . MINOR . MICRO.

Ah, GStreamer provides nano version.

I added nano to `Gst::Version::STRING` and `Gst::Version.or_later?` after thinking.
Because...
`Gst::Version::STRING`: Users will want to know about full version information.
`Gst::Version.or_later?`: We can provide the same API as other `or_later?` family. It means that we can call `Gst::Version.or_later?` with only major and minor versions such as `Gst::Version.or_later?(1, 4)`.

cedlemo:
Ok,
if I found another ruby-gnome2 that provides NANO, I will do the same.

PR #580, opened 2015-09-26
Update TODO

cedlemo:
The new gtk3 version is out 3.18, good news there are no new demos since the last check. But the main part has been rewriten using `Gtk::Application`. Should we do the same? (I would like because I thnk we should stay up to date)

### DIFF
--- a/gtk3/sample/gtk-demo/TODO
+++ b/gtk3/sample/gtk-demo/TODO
@@ -1,4 +1,4 @@
-# C version (3.17.6) versus Ruby version
+# C version (3.18.0) versus Ruby version
 
 C version             Ruby version    Updated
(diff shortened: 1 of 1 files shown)


kou:
> Should we do the same?

I think so!

cedlemo:
How should I proceed?
create a branch in my local repository and push when all the demos are finished or can I push step by step (assuming that what works now will not works when  the main file will be modified).

kou:
I like "step by step" approach.
If you find a problem, I want to fix it step by step.

> (assuming that what works now will not works when the main file will be modified)

It's OK that components that are updated yet don't work while we'are updating gtk-demo.

cedlemo:
ok

PR #586, opened 2015-10-02
Gtk3 sample gtk demo main rb interface

cedlemo:
*    add all data files needed
*    modify the gresource xml file in order to not list the non impelemented source files
*    start implementing the activate method

I will wait a review from you before going further.


kou:
Thanks.

I merged and removed `#activate`. Because `#activate` is defined in `Gio::Application`. We should not overwrite it.

See https://github.com/ruby-gnome2/ruby-gnome2/commit/738ac547370284637571aa36a6d9428ab1c45922 for details.

cedlemo:
Ok,

I just tryed to follow the main.c source file. Now, I am a little lost in the main.c file, the "comand-line" event triggers the comand_line cb that calls the *activate* function. But  this function is also the  callback to the "activate" event. How should I reproduce this?

kou:
Yes. It's `run_application`.

PR #588, opened 2015-10-03
Gtk3 sample gtk demo main rb run application

cedlemo:
I create this pull request because the next step should be to list the demos and populate the model of the treeview


kou:
Thanks.
I've changed to use `Gtk::Builder#[]` instead of `#get_object`. `#[]` is more Rubyish.

PR #589, opened 2015-10-04
Add list of demos

cedlemo:
This is an adaptation of the models.h file that is generated by `geninclude.pl`. This is used to populate the treeview model. I am not sure it is the best way to store the demos list and informations.

We must anticipate/ decide on the model we want for the demo files and on the way to launch the demo. 

In C it seems that the demos sources contains both code for lib and for an executable. The main.c file call a function like `do_application` for example that launch an executable *gtk-demo-application* which is created using the code contained  in the `#ifdef STANDALONE ... #endif` macro.
*edit* 
This example is just for application.c

I really need your help in order to know how you want to implement this.

Regards


kou:
Sorry. I missed it.

I prefer to extract the information from *.rb files dynamically. We did so before you rewrite gtk-demo: https://github.com/ruby-gnome2/ruby-gnome2/blob/master/gtk3-no-gi/sample/gtk-demo/main.rb#L76-L103

cedlemo:
Oh yes!!

cedlemo:
@kou, 

This should be ok now.

kou:
Good!

PR #593, opened 2015-10-14
Gtk3 sample gtk demo main rb list demos method

cedlemo:
There is an important fix in this pull request, when I simplified the code, the children where not well registred.

The list_demos method works when you launch:

    ./main.rb --list


kou:
Thanks!

PR #595, opened 2015-10-17
add a test module and implement the use of test_mod.rb

cedlemo:
The test_mod.rb is only for the test and implementation purpose. It will be removed later (It just help me to focus on the way to load the demo files and code.


kou:
OK!

PR #597, opened 2015-10-17
Gtk3 sample gtk demo main rb run demo for module

cedlemo:
After that it will remain to implement:

*    auto_quit callback
*    the callback for treeview selection-changed signal
*    activate_about callback
*    activate_quit callback
*    selection_cb callback
*    file loading function
*    basic syntax highlighting


kou:
OK!

PR #607, opened 2015-10-22
Gtk3 sample gtk demo main rb load file

cedlemo:
It remains the `add_data_tab` to implement, but the main.rb is fully working .


kou:
Thanks. Could you remove needless log file?

Sorry for my late response. I'll be busy until RubyKaigi 2015 is finished... (Dec. 11th - 13th)

cedlemo:
log file removed.

> I'll be busy until RubyKaigi 2015 is finished... (Dec. 11th - 13th)

If you can at least merge this one because I plan to start updating the demo modules. It remains, in the main.rb, the functionality that displays demo data in tab but I can implement it later.

Thanks.

Issue #608, opened 2015-10-28
Gtk::Widget#destroyed


cedlemo:
This method which should be used like this ( I though):
    
```ruby
mywidget.destroyed 
# or
Gtk::Widget.destroyed(mywidget)
```

Can only be use like this:

```ruby
mywidget.destroyed(mywidget)
```

*   api https://developer.gnome.org/gtk3/unstable/GtkWidget.html#gtk-widget-destroyed
*   usage : https://git.gnome.org/browse/gtk+/tree/demos/gtk-demo/scale.c#n25

```c
     g_signal_connect (window, "destroy",
                        G_CALLBACK (gtk_widget_destroyed), &window);
```

The ruby adaptation:

```ruby
    window.signal_connect("destroy") { window.destroyed(window) }
```

I don't even know if this function is needed with the ruby bindings.

kou:
We don't need to use `gtk_widget_destroyed()` because it just sets `NULL` to variable for widget pointer in C. We don't need to manage `NULL` In Ruby.

PR #609, opened 2015-10-28
Gtk3 sample gtk demo scale rb

cedlemo:
See issue #608

### DIFF
--- a/gtk3/sample/gtk-demo/scale.rb
+++ b/gtk3/sample/gtk-demo/scale.rb
@@ -1,0 +1,28 @@
+# Copyright (c) 2015 Ruby-GNOME2 Project Team
+# This program is licenced under the same licence as Ruby-GNOME2.
+#
+=begin
+= Scale
+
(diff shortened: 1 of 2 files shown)


cedlemo:
Rebase done 7-nov-2015

gtk3/sample/gtk-demo/scale.rb
```diff
@@ -0,0 +1,28 @@
...
+=end
+
+module ScaleDemo
+  def self.run_demo(main_window)
+    builder = Gtk::Builder.new(:resource => "/scale/scale.ui")
+    builder.connect_signals {}
+    window = builder["window1"]
+    window.screen = main_window.screen
+
+    window.signal_connect("destroy") { window.destroyed(window) }
```
kou: You can just remove this line. Because it's only for C. It's not needed for Ruby.

cedlemo:
Line removed.

kou:
Thanks!

PR #613, opened 2015-11-04
Finish adaptation of the official tutorial

cedlemo:


kou:
Thanks!
(Replying to ruby-gnome2-devel-en is also very helpful! Thanks!)

cedlemo:
> Replying to ruby-gnome2-devel-en is also very helpful!

I try ! :)

PR #618, opened 2015-11-11
Gtk3 sample gtk demo entry buffer rb

cedlemo:


cedlemo:
The #617 should be merged before. (`Gtk::EntryBuffer.new` usage)

kou:
Thanks!!!

PR #625, opened 2015-11-20
Update links demo

cedlemo:


kou:
Thanks!!!
But it has some extra new lines in texts.

PR #630, opened 2015-11-23
Gtk3 sample gtk demo entry completion rb

cedlemo:


cedlemo:
Was not created on upstream/master so I delete it

PR #632, opened 2015-11-25
Add revealer demo

cedlemo:
Sorry for the name of the branch (it should be gtk3_sample_gtk_demo_revealer_rb instead of gtk3_sample_gtk_demo_entry_revealer_rb)


kou:
Thanks!

PR #633, opened 2015-11-25
Update expander demo

cedlemo:


gtk3/sample/gtk-demo/expander.rb
```diff
@@ -1,42 +1,75 @@
...
-      vbox.border_width = 5
+    tv = Gtk::TextView.new
+    tv.set_editable(false)
+    tv.set_wrap_mode(:word)
+    tv.buffer.text = <<TEXT
+Finally, the full story with all details.
+And all the inside information, including
+error codes, etc etc. Pages of information,
+you might have to scroll down to read it all,
+or even resize the window - it works !\n
```
kou: Is `\n` needed? This is a here document.

cedlemo:
*    I think that a line between the two paragraphs is better (visualy)
*    but I think that in a here document we should use directly a new line instead of `\n`

So no the `\n` is not needed

kou:
Thanks. I've merged!

PR #636, opened 2015-11-28
Update colorsel demo

cedlemo:
In the line 52, I added `da.queue_draw` because when I tested the demo, I noticed this each time:

*    launch the demo
*    change the color --> nothing happens in the drawing area
*    change the color --> drawing area with a new background color

With this I am sure that the change are effective.


kou:
Thanks!!!

PR #637, opened 2015-11-30
Update css_basics demo

cedlemo:


gtk3/sample/gtk-demo/css_basics.rb
```diff
@@ -1,103 +1,76 @@
...
-        css_text_changed(_text, provider)
-      end
-
-      text.text = File.read(File.join(__dir__, "css_basics.css"))
-
-      provider.signal_connect("parsing-error") do |_provider, section, error|
-        p section
-        p error
-        show_parsing_error(section, error, child.buffer)
+def apply_style(widget, provider)
```
kou: This method should be defined in `CssBasicDemo` because this method is only used in `CssBasicDemo`.

cedlemo: Ok,
so I guess I have to do the same for the css_accordion.rb demo I have done previously :-)
https://github.com/ruby-gnome2/ruby-gnome2/blob/master/gtk3/sample/gtk-demo/css_accordion.rb

kou: > so I guess I have to do the same for the css_accordion.rb demo I have done previously :-)

Yes, please!

cedlemo:
Done

kou:
Thanks!!!!

PR #638, opened 2015-12-01
"rbgobj_gc_mark_instance" has to be called for the GtkTreeSelection

htrb:
"rbgobj_gc_mark_instance" has to be called for the GtkTreeSelection associated with tree_view.


kou:
Thanks!

PR #641, opened 2015-12-03
add support for EntryCompletionMatchFunc.

htrb:


gtk3/lib/gtk3/container.rb
```diff
@@ -36,11 +36,11 @@ def add(child, properties={})
       ensure
         child.thaw_child_notify
       end
+      self
     end
 
     def <<(widget)
```
kou: `<<` should be just an alias.

htrb:
add support for EntryCompletionMatchFunc.

PR #644, opened 2015-12-05
Add support for GtkTreeSelection

dutchhome:
This is a tested working TreeSelectionFunc implementation for Gtk3.  I have not included a test case as I'm not sure the best way to test this without opening a window.


kou:
Thanks!

Issue #645, opened 2015-12-05
Css doesn't apply to the GtkWindow


cedlemo:
I have updated the demo css_pixbufs and I have no error except the fact that I can not display the images or the animations on the window with the default css .

if I replace the css selector `GtkWindow` with the generic one `*{}`, I am able to display the images but I can not see any animations.

Any Ideas ?

kou:
[GtkInspector](https://wiki.gnome.org/Projects/GTK%2B/Inspector) may help you.

```text
% GTK_DEBUG=interactive gtk3/sample/gtk-demo/main.rb
```

cedlemo:
I will use it next time. Thx

PR #646, opened 2015-12-05
Update css_pixbufs demo

cedlemo:
Please don't merge right now have a look at this issue #645

remains:

*    fix pb with css
*    update TODO file

### DIFF
--- a/gtk3/sample/gtk-demo/css_pixbufs.rb
+++ b/gtk3/sample/gtk-demo/css_pixbufs.rb
@@ -1,0 +1,83 @@
+# Copyright (c) 2015 Ruby-GNOME2 Project Team
+# This program is licenced under the same licence as Ruby-GNOME2.
+#
+=begin
+=  CSS Theming/Animated Backgrounds
+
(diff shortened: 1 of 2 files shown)


gtk3/sample/gtk-demo/css_pixbufs.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    if !window.visible?
+      window.show_all
+    else
+      window.destroy
+    end
+    window
+  end
+
+  def self.apply_style(widget, provider)
```
kou: Change to the following:

```ruby
  def self.apply_style(widget, provider)
    style_context = widget.style_context
    style_context.add_provider(provider, Gtk::StyleProvider::PRIORITY_USER)
    widget.children.each do |child|
      apply_style(child, provider)
    end
  end
```

cedlemo: Wow, awesome. As always Thanks!.

This function is present in other demo and I see if I have to modify those files.

kou:
Thanks!

PR #651, opened 2015-12-21
add support for EntryCompletionMatchFunc

htrb:
add support for EntryCompletionMatchFunc


kou:
Could you rebase on master and remove needless commits?

PR #654, opened 2015-12-25
Make Gtk::Widget#translate_coordinates more rubyish

cedlemo:
`Gtk::Widget#translate_coordinates` returns an array like this [x, y] if it works or nil.


kou:
Thanks!

PR #664, opened 2015-12-31
Update search_entry2 demo

cedlemo:


gtk3/sample/gtk-demo/search_entry2.rb
```diff
@@ -0,0 +1,107 @@
...
+
+ GtkSearchEntry sets up GtkEntries ready for search. Search entries
+ have their "changed" signal delayed and should be used
+ when the searched operation is slow such as loads of entries
+ to search, or online searches.
+=end
+module SearchEntry2Demo
+  def self.run_demo(main_window)
+    window = Gtk::Window.new(:toplevel)
+    window.set_title("Delayed Search Entry")
```
kou: Could you use assign form such as `window.title = "Delayed Search Entry" rather than setter form?
Because assign form is more Rubyish rather than setter form.

cedlemo:
Done.

kou:
Thanks!!!

PR #673, opened 2016-01-13
glib2: support more variant types in rbg_variant_to_ruby()

aurelj:


glib2/ext/glib2/rbglib-variant.c
```diff
@@ -39,11 +39,39 @@ rbg_variant_to_ruby(GVariant *variant)
...
         return CSTR2RVAL_LEN(string, string_length);
+    } else if (g_variant_type_equal(type, G_VARIANT_TYPE_HANDLE)) {
+        return INT2NUM(g_variant_get_handle(variant));
+    } else if (g_variant_type_equal(type, G_VARIANT_TYPE_VARIANT)) {
+        return rbg_variant_new(g_variant_get_variant(variant));
+    } else if (g_variant_type_is_array(type)) {
+        gsize len = g_variant_n_children(variant);
+        VALUE ary = rb_ary_new2(len);
+        for (gsize i = 0; i < len; i++)
+            rb_ary_store(ary, i, rbg_variant_to_ruby(g_variant_get_child_value(variant, i)));
```
kou: We should call `g_variant_unref()` for return value of `g_variant_get_child_value()`.

https://developer.gnome.org/glib/stable/glib-GVariant.html#g-variant-get-child-value

> The returned value is never floating. You should free it with `g_variant_unref()` when you're done with it.

glib2/ext/glib2/rbglib-variant.c
```diff
@@ -39,11 +39,39 @@ rbg_variant_to_ruby(GVariant *variant)
...
         string = g_variant_get_string(variant, &string_length);
         return CSTR2RVAL_LEN(string, string_length);
+    } else if (g_variant_type_equal(type, G_VARIANT_TYPE_HANDLE)) {
+        return INT2NUM(g_variant_get_handle(variant));
+    } else if (g_variant_type_equal(type, G_VARIANT_TYPE_VARIANT)) {
+        return rbg_variant_new(g_variant_get_variant(variant));
+    } else if (g_variant_type_is_array(type)) {
+        gsize len = g_variant_n_children(variant);
+        VALUE ary = rb_ary_new2(len);
+        for (gsize i = 0; i < len; i++)
```
kou: Don't define variable in `for`.

glib2/ext/glib2/rbglib-variant.c
```diff
@@ -39,11 +39,39 @@ rbg_variant_to_ruby(GVariant *variant)
...
+               g_variant_type_equal(type, G_VARIANT_TYPE_OBJECT_PATH) ||
+               g_variant_type_equal(type, G_VARIANT_TYPE_SIGNATURE)) {
         const gchar *string;
         gsize string_length;
         string = g_variant_get_string(variant, &string_length);
         return CSTR2RVAL_LEN(string, string_length);
+    } else if (g_variant_type_equal(type, G_VARIANT_TYPE_HANDLE)) {
+        return INT2NUM(g_variant_get_handle(variant));
+    } else if (g_variant_type_equal(type, G_VARIANT_TYPE_VARIANT)) {
+        return rbg_variant_new(g_variant_get_variant(variant));
```
kou: We should call `g_variant_unref()` for return value of `g_variant_get_variant()`.

glib2/ext/glib2/rbglib-variant.c
aurelj: On Thu, Jan 14, 2016 at 02:49:06AM -0800, Kouhei Sutou wrote:
> >          const gchar *string;
> >          gsize string_length;
> >          string = g_variant_get_string(variant, &string_length);
> >          return CSTR2RVAL_LEN(string, string_length);
> > +    } else if (g_variant_type_equal(type, G_VARIANT_TYPE_HANDLE)) {
> > +        return INT2NUM(g_variant_get_handle(variant));
> > +    } else if (g_variant_type_equal(type, G_VARIANT_TYPE_VARIANT)) {
> > +        return rbg_variant_new(g_variant_get_variant(variant));
> > +    } else if (g_variant_type_is_array(type)) {
> > +        gsize len = g_variant_n_children(variant);
> > +        VALUE ary = rb_ary_new2(len);
> > +        for (gsize i = 0; i < len; i++)
> > +            rb_ary_store(ary, i, rbg_variant_to_ruby(g_variant_get_child_value(variant, i)));
> 
> We should call `g_variant_unref()` for return value of `g_variant_get_child_value()`.

Good catch, thanks, fixed (and the other one as well).

glib2/ext/glib2/rbglib-variant.c
aurelj: On Thu, Jan 14, 2016 at 02:49:18AM -0800, Kouhei Sutou wrote:
> >          const gchar *string;
> >          gsize string_length;
> >          string = g_variant_get_string(variant, &string_length);
> >          return CSTR2RVAL_LEN(string, string_length);
> > +    } else if (g_variant_type_equal(type, G_VARIANT_TYPE_HANDLE)) {
> > +        return INT2NUM(g_variant_get_handle(variant));
> > +    } else if (g_variant_type_equal(type, G_VARIANT_TYPE_VARIANT)) {
> > +        return rbg_variant_new(g_variant_get_variant(variant));
> > +    } else if (g_variant_type_is_array(type)) {
> > +        gsize len = g_variant_n_children(variant);
> > +        VALUE ary = rb_ary_new2(len);
> > +        for (gsize i = 0; i < len; i++)
> 
> Don't define variable in `for`.

OK, fixed.
But it is a bit sad to explicitely avoid some nice language features
that are part of the ISO C standard since more than 17 years now ! (C99)

Moreover, now that gcc5 defaults to C11 (while gcc <= 4.9 defaults to
C90), we should probably expicitely specify in the build system what
version of the C standard we are using. Without this, the code is
currently compiled following different standard depending on the compiler
used.
I would suggest to add a -std=c11 to the compiler options. (This is
supported since gcc 4.7 which is already 4 years old)

kou: We still support CentOS 5 that was released 9 years ago. :<

Anyway, it may be good time to require C99.

kou:
Thanks.
I've merged

PR #675, opened 2016-01-13
glib2: fix handling of 64 bits FIXNUM variant initialization

aurelj:
This fixes the following error:

irb(main):001:0> require 'glib2'
=> true
irb(main):002:0> GLib::Variant.new(333222111000)
RangeError: integer 333222111000 too big to convert to `int'
            from (irb):2:in `initialize'
            from (irb):2:in `new'
            from (irb):2
            from /usr/bin/irb:11:in `<main>'

### DIFF
--- a/glib2/ext/glib2/rbglib-variant.c
+++ b/glib2/ext/glib2/rbglib-variant.c
@@ -95,7 +95,7 @@ rg_initialize(int argc, VALUE *argv, VALUE self)
             variant_type = G_VARIANT_TYPE_BOOLEAN;
             break;
           case RUBY_T_FIXNUM:
-            variant_type = G_VARIANT_TYPE_INT32;
+            variant_type = G_VARIANT_TYPE_INT64;
             break;
           case RUBY_T_FLOAT:
             variant_type = G_VARIANT_TYPE_DOUBLE;


kou:
Thanks for the pull request.
But is it OK that we treat all integers as 64bit integer? Or should we do integer range check to choose 32bit integer or 64bit integer?
I don't know about it. Do you have any knowledge about this case?

aurelj:
On Thu, Jan 14, 2016 at 03:02:54AM -0800, Kouhei Sutou wrote:
> Thanks for the pull request.
> But is it OK that we treat all integers as 64bit integer? Or should we do integer range check to choose 32bit integer or 64bit integer?
> I don't know about it. Do you have any knowledge about this case?

We should use a GVariantType that can fit any value of FIXNUM.
Choosing the GVariantType depending on a range check would be less user
friendly I think, as user couldn't rely on one specific type for any
integer.

The only drawback there is to using a 64 bit integer is a small
performance hit on 32 bits systems and a small memory usage increase.
But nobody will ever use GVariant (even less thru Ruby) on some
performance critical code path so I think it really doesn't matter
in this case.

kou:
OK. I've merged it.
If someone reports a problem case, we'll consider about this again.

PR #676, opened 2016-01-14
Implement GLib::Regex#split methods and tests

cedlemo:


glib2/ext/glib2/rbglib_regex.c
```diff
@@ -76,6 +76,52 @@ rg_match_flags(VALUE self)
...
+    const gchar *string;
+    gssize string_len = -1;
+    gint start_position = 0;
+    GRegexMatchFlags match_options = 0;
+    gint max_tokens = 0;
+
+    rb_scan_args(argc, argv, "11", &rb_string, &rb_options);
+
+    rbg_scan_options(rb_options,
+                     "string_len", &rb_string_len,
```
kou: We should use `RSTRING_LEN(rb_string)` for `string_len` instead of specifying as an option.

cedlemo: Ok done!

What will be the next steps?

kou:
Thanks!
How about "GLib::Rexgex#match" and `GLib::MatchInfo#matches?` as the next step?

cedlemo:
@kou 
Done #678 sorry for the delay. I will have more time this week. I wait for the next instructions.

kou:
> Done #678 sorry for the delay.

No problem. :-)

> I wait for the next instructions.

Could you resolve the TODOs?:

* https://github.com/ruby-gnome2/ruby-gnome2/blob/master/glib2/test/test-regex.rb#L84
* https://github.com/ruby-gnome2/ruby-gnome2/blob/master/glib2/test/test-regex.rb#L90

And could you define `(GMatchInfo*)RVAL2BOXED(s, G_TYPE_MATCH_INFO)` at https://github.com/ruby-gnome2/ruby-gnome2/blob/master/glib2/ext/glib2/rbglib_matchinfo.c#L23 in https://github.com/ruby-gnome2/ruby-gnome2/blob/master/glib2/ext/glib2/rbglib2conversions.h and use it in `*.c`?

PR #678, opened 2016-01-18
Glib2 regex match

cedlemo:
*    `GLib::MatchInfo` class
*    `GLib::MatchInfo#string`
*    `GLib::MatchInfo#regex`
*    `GLib::MatchInfo#matches?`
*    `GLib::Regex#match`
*    and tests

### DIFF
--- a/glib2/ext/glib2/rbglib_matchinfo.c
+++ b/glib2/ext/glib2/rbglib_matchinfo.c
@@ -1,0 +1,39 @@
+/*
+ *  Copyright (C) 2015  Ruby-GNOME2 Project Team
+ *
+ *  This library is free software; you can redistribute it and/or
+ *  modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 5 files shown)


glib2/ext/glib2/rbglib_matchinfo.c
```diff
@@ -0,0 +1,63 @@
...
+ *  MA  02110-1301  USA
+ */
+
+#include "rbgprivate.h"
+
+#define RG_TARGET_NAMESPACE cMatchInfo
+#define _SELF(s) ((GMatchInfo*)RVAL2BOXED(s, G_TYPE_MATCH_INFO))
+
+static void
+match_info_free(gpointer object)
```
kou: This is not needed.
It's done automatically by Ruby/GLib2. :-)

kou:
Thanks!

PR #679, opened 2016-01-18
Gtk3 sample gtk demo dialog rb

cedlemo:


gtk3/sample/gtk-demo/dialog.rb
```diff
@@ -1,137 +1,127 @@
...
+    hbox.pack_start(vbox2, :expand => false, :fill => false, :padding => 0)
+    vbox2.pack_start(button, :expand => false, :fill => false, :padding => 0)
+
+    table = Gtk::Grid.new
+    table.row_spacing = 4
+    table.column_spacing = 4
+    hbox.pack_start(table, :expand => false, :fill => false, :padding => 0)
+
+    label = Gtk::Label.new("_Entry 1", :use_underline => true)
+    table.attach(label, 0, 0, 1, 1)
```
kou: I'll merge this pull request but I add a comment.
We should improve `Gtk::Grid#attach` API for easy to use. For example:

```ruby
def attach(label, options={})
  left   = options[:left]   || 0
  top    = options[:top]    || 0
  width  = options[:width]  || 1
  height = options[:height] || 1
  attach_raw(label, left, top, width, height)
end
```

kou:
Thanks!

gtk3/sample/gtk-demo/dialog.rb
cedlemo: @kou 
Tell me if you want me to do it

kou: Thanks. Could you try to do it? I want to see scripts that uses the API to confirm the API is really easy to use.

PR #680, opened 2016-01-19
Gtk3 lib grid attach

cedlemo:
I have done as you asked me. You can see in the test one example.


kou:
Thanks!
I'll look at it later.

PR #681, opened 2016-01-19
Glib2 match info test and conversions

cedlemo:
>Could you resolve the TODOs?:
https://github.com/ruby-gnome2/ruby-gnome2/blob/master/glib2/test/test-regex.rb#L84
https://github.com/ruby-gnome2/ruby-gnome2/blob/master/glib2/test/test-regex.rb#L90

Yes, it is done.

>And could you define (GMatchInfo*)RVAL2BOXED(s, G_TYPE_MATCH_INFO) at https://github.com/ruby-gnome2/ruby-gnome2/blob/master/glib2/ext/glib2/rbglib_matchinfo.c#L23 in https://github.com/ruby-gnome2/ruby-gnome2/blob/master/glib2/ext/glib2/rbglib2conversions.h and use it in *.c?

I have interpreted what you have written instead of doing litteraly what you asked me and I have created the macro `GMATCHINFO2RVAL` because `(GMatchInfo*)RVAL2BOXED(s, G_TYPE_MATCH_INFO)` this part of code is only needed in the rbglib_matchinfo.c file and the `_SELF` macro usage seems to be the standard. Is it OK?


kou:
Thanks!

> I have interpreted what you have written instead of doing litteraly what you asked me and I have created the macro `GMATCHINFO2RVAL` because `(GMatchInfo*)RVAL2BOXED(s, G_TYPE_MATCH_INFO)` this part of code is only needed in the rbglib_matchinfo.c file and the `_SELF` macro usage seems to be the standard. Is it OK?

It's for libraries that depend on glib2 gem. If all libraries are based on GObject Introspection, we don't need the header.

cedlemo:
@kou what are the next steps?

kou:
Sorry for no indication.
Could you send pull requests that has only one method (or a few related methods such as `Regex#match` and `MatchInfo` case)? It's easy to review when one pull request have only one method.

(The following English may be broken. Sorry.)
I don't specify concrete order for target methods. Please implement easier methods then difficult methods. For example, `Regex.escape_string` is the first and `Regex#replace` is the last.

I'm sorry but I'll be busy until 2016-02-17. I have some talks in some events...
I'll review your pull request but it may be slow.

cedlemo:
>I'm sorry but I'll be busy until 2016-02-17. I have some talks in some events...

No problem Bon courage! (good luck or がんばください I guess)

kou:
Thanks!
(がんばってください is correct :-)

PR #686, opened 2016-01-23
Glib2 regex capture count

cedlemo:
This one should be OK


kou:
Ah, sorry.
Could you rebase on master? #684 conflicts with it.

cedlemo:
Yes of course.
For each pull request I create them using:

    git checkout -b mybranch upstream/master

Is there a better way that make the thing easier for you?

kou:
Use the following!:

```text
git checkout mybranch
git fetch --all
git rebase upstream/master
```

kou:
Thanks!

PR #688, opened 2016-01-23
Fix missing return type for rbg_capture_count

cedlemo:
Fix for my error in  6a2df50


cedlemo:
Not needed , fixed in #687

Issue #682, opened 2016-01-23
Implement Gtk::TreeIter#first_child


kou:
https://sourceforge.net/p/ruby-gnome2/mailman/message/34788222/

It should be `iter.nth_child(0)`.

cedlemo:
@kou ,

Maybe you could close this one

kou:
Good catch!

PR #691, opened 2016-01-25
Improve and finish tests for Gtk::TreeIter

cedlemo:
Before working on #682 I think we should finish and improve the tests


kou:
Thanks!!

PR #692, opened 2016-01-25
Add another example for Gtk::TreeStore

cedlemo:
It already exist a sample for the `Gtk::TreeView` that uses `Gtk::TreeStore` but this one is more focused on the `Gtk::TreeStore` and more simple.


kou:
Thanks!!!

PR #693, opened 2016-01-25
Gtk3 lib tree iter first child

cedlemo:
for issue #682

### DIFF
--- a/gtk3/lib/gtk3/tree-iter.rb
+++ b/gtk3/lib/gtk3/tree-iter.rb
@@ -1,4 +1,4 @@
-# Copyright (C) 2014-2015  Ruby-GNOME2 Project Team
+# Copyright (C) 2014-2016 Ruby-GNOME2 Project Team
 #
 # This library is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
@@ -71,6 +71,10 @@ module Gtk
 
     def children
       @model.iter_children(self)
+    end
+
+    def first_child
+      @model.iter_nth_child(self, 0)
     end
 
     def ==(other)
--- a/gtk3/lib/gtk3/tree-iter.rb
+++ b/gtk3/lib/gtk3/tree-iter.rb
@@ -74,7 +74,7 @@ module Gtk
     end
 
     def first_child
-      @model.iter_nth_child(self, 0)
+      nth_child(0)
     end
 
     def ==(other)
--- a/gtk3/test/test-gtk-tree-iter.rb
+++ b/gtk3/test/test-gtk-tree-iter.rb
@@ -102,6 +102,15 @@ class TestGtkTreeIter < Test::Unit::TestCase
     assert_equal(second_child_iter, @iter.nth_child(1))
   end
 
+  def test_firs_child
+    @iter.values = ["Dad"]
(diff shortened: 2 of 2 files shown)


gtk3/lib/gtk3/tree-iter.rb
```diff
@@ -73,6 +73,10 @@ def children
       @model.iter_children(self)
     end
 
+    def first_child
+      @model.iter_nth_child(self, 0)
```
kou: Could you use `#nth_child` to reduce the same code?

gtk3/test/test-gtk-tree-iter.rb
```diff
@@ -102,6 +102,15 @@ def test_nth_children
     assert_equal(second_child_iter, @iter.nth_child(1))
   end
 
+  def test_firs_child
+    @iter.values = ["Dad"]
+    first_child_iter = @model.append(@iter)
+    first_child_iter.values = ["First son"]
+    second_child_iter = @model.append(@iter)
+    second_child_iter.values = ["First daughter"]
+    assert_equal(@iter.nth_child(0), @iter.first_child)
```
kou: `first_child_iter` is better rather than `@iter.nth_child(0)` because we know `first_child_iter` is the first child in this context.

gtk3/lib/gtk3/tree-iter.rb
cedlemo: Yes of course I will but I have one question, I thought (or at least I had the feeling) that how I did it was more efficient, look:

```
def first_child
   @model.iter_nth_child(self, 0)
end
```

*    ruby get the instance variable value
*    then ask its class `Gtk::TreeStore` for the method `#iter_nth_child` and so on.

And for me the usage of the `#nth_child` method we have defined before, add another steps : 

```
def first_child
    nth_child(0)
end
```
*   get self class, 
*   ask for the method nth_child
*    ruby get the instance variable value
*    then ask its class `Gtk::TreeStore` for the method `#iter_nth_child` and so on.

I well understand that this example taken alone is not sufficient to assure a better efficiency of the code but I though that in complexe code the cumulated efficiency of the code could matter.

I have a naive/basic knowledge of ruby and maybe it must have some magic code behind the ruby interpreter that make that the two previous codes are the same.

Or maybe you think that the readability if more important thant the efficiency we could add?

As I have said before I will do what you asked me to do, I just try to have a better understanding of what I am doing.

kou: Yes. Your implementation is more efficient. There is no magic.

> Or maybe you think that the readability if more important that the efficiency we could add?

Yes. I choose maintainability and readability in this case.

With your implementation, if we need to change `#nth_child` implementation, we also need to change `#first_child` implementation. (maintainability)

The current implementation says "the first child is 0th child". It's straightforward. (readability)

> As I have said before I will do what you asked me to do, I just try to have a better understanding of what I am doing.

I like your style. :-)
If you have any question in the future, please ask me as usual. :-)

kou:
Thanks!

gtk3/lib/gtk3/tree-iter.rb
cedlemo: Ok Thanks

Issue #683, opened 2016-01-23
Add missing ! to Gtk::TreePath#next, #prev, #up and #down


kou:
https://sourceforge.net/p/ruby-gnome2/mailman/message/34788222/

It should be done in `Gtk::Loader#load_method_info`.

kou:
Done by #694.
(Adding tests is better.)

cedlemo:
>(Adding tests is better.)

Message received :-) (I will do them today or tomorrow)

kou:
Wow! Thanks!

PR #694, opened 2016-01-25
Add Gtk::TreePath#next! , prev!, up! and down! methods

cedlemo:
For the issue #683. 
@kou I hope this is the best to do it.


kou:
Thanks!
It's the best!

PR #695, opened 2016-01-25
Glib2 regex escape string

cedlemo:


glib2/test/test-regex.rb
```diff
@@ -190,4 +190,8 @@ def test_match_flags
       assert_equal(2, regex.string_number("a_name"))
     end
   end
+
+  def test_regex_escape_string
+    assert_equal(GLib::Regex.escape_string("a.b*c"), "a\\.b\\*c")
```
kou: The first argument of `assert_equal` is the expected value because the expected value is more important for readers who read test code:

```ruby
assert_equal("a\\.b\\*c", GLib::Regex.escape_string("a.b*c"))
```

cedlemo: Oups Sorry,
I will be more carefull on that

cedlemo:
Rebase done

kou:
Thanks!!

PR #697, opened 2016-01-28
Glib2 regex match simple

cedlemo:


kou:
How about implementing `GLib::Regex.match` by the following?:

```ruby
class << self
  def match(pattern, string, options={})
    new(pattern, options).match(string, options)
  end
end
```

I don't want to use `_simple` suffix in method name. Because it's not Rubyish. Rubyish API don't provide simple feature method and full feature method. You can use simple feature by not providing additional argument.

cedlemo:
Of course, I didn't thought about that.

cedlemo:
Done

glib2/lib/glib2/regex.rb
```diff
@@ -0,0 +1,25 @@
...
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+module GLib
+  class Regex
+    class << self
+      def match(pattern, string, options={})
```
kou: Ah, it should have `?` suffix. Because it's a predicate method.

glib2/test/test-regex.rb
```diff
@@ -194,4 +194,15 @@ def test_match_flags
   def test_regex_escape_string
     assert_equal("a\\.b\\*c", GLib::Regex.escape_string("a.b*c"))
   end
+
+  sub_test_case "match_simple" do
+    test "true" do
+      matched = GLib::Regex.match("to", "tatota")
+      assert(matched)
```
kou: Use the following style:

```ruby
assert do
  GLib::Regex.match?("to", "tatota")
end
```

cedlemo:
Done :)

kou:
Thanks!

PR #699, opened 2016-02-05
Glib2 regex match all

cedlemo:
For the next pull requests: 

*    Should I have to implement `GLib::Regex.split` ( ``g_regex_split_simple`) in ruby using the method `GLib::Regex#split` like we did previously for `g_regex_match_simple`.


kou:
Thanks!

cedlemo:
@kou ,

> For the next pull requests:

>Should I have to implement GLib::Regex.split ( `g_regex_split_simple) in ruby using the method GLib::Regex#split like we did previously for g_regex_match_simple.

any hints?

kou:
Sorry. I missed it.
It looks good to me as the next pull request.

PR #704, opened 2016-02-10
Implement GLib::Regex#replace

cedlemo:


glib2/ext/glib2/rbglib_regex.c
```diff
@@ -260,6 +260,47 @@ rg_match_all(gint argc, VALUE *argv, VALUE self)
...
+                                      string,
+                                      string_len,
+                                      start_position,
+                                      replacement,
+                                      match_options,
+                                      &error);
+    if (error)
+        RAISE_GERROR(error);
+
+    return CSTR2RVAL(modified_string);
```
kou: Use `CSTR2RVAL_FREE()`. The current code leaks a memory.

glib2/test/test-regex.rb
```diff
@@ -250,4 +250,16 @@ def test_regex_escape_string
       assert_equal(["toto ", ""], splited_strings)
     end
   end
+
+  sub_test_case "replace" do
+    test "simple" do
+      regex = GLib::Regex.new("\s")
```
kou: It should be `"\\s"`.

cedlemo:
Modifications done.

kou:
Thanks!

PR #710, opened 2016-02-11
Glib2 regex replace literal

cedlemo:


kou:
How about adding `:literal => true` option to `GLib::Regex#replace`?

cedlemo:
Yes of course. It is way better.

kou:
Thanks!

PR #712, opened 2016-02-12
Glib2 regex check replacement

cedlemo:
Maybe you won't agree with this. I wait for your advices.

It remains to do the `GLib::Regex#replace_eval` which will be tricky. Do you want that I do the simple methods for `GLib::MacthInfo` before it?

If you want that I do the `GLib::Regex#replace_eval ` in the next step you should look here https://github.com/cedlemo/ruby-gnome2/blob/f38f53d2032e0df457d88a2c2d1a4f2f65b2d9e9/glib2/ext/glib2/rbglib_regex.c#L212 and https://github.com/cedlemo/ruby-gnome2/blob/f38f53d2032e0df457d88a2c2d1a4f2f65b2d9e9/glib2/ext/glib2/rbglib_regex.c#L254 . This is How I plan to implement it. Tell me if there is something wrong with this.

### DIFF
--- a/glib2/ext/glib2/rbglib_regex.c
+++ b/glib2/ext/glib2/rbglib_regex.c
@@ -325,6 +325,29 @@ rg_s_escape_string(G_GNUC_UNUSED VALUE self, VALUE string)
     return CSTR2RVAL(g_regex_escape_string(RVAL2CSTR(string), RSTRING_LEN(string)));
 }
 
+  static VALUE
+rg_s_check_replacement(G_GNUC_UNUSED VALUE self, VALUE rb_replacement)
+{
+    gboolean is_valid, has_ref;
+    const gchar * replacement = RVAL2CSTR(rb_replacement);
+    GError *error = NULL;
+    VALUE boolean_val;
+
+    is_valid = g_regex_check_replacement(replacement, &has_ref, &error);
+
+    if (error)
+        RAISE_GERROR(error);
+
+    VALUE ret = rb_ary_new();
+
+    boolean_val = is_valid == TRUE ? Qtrue : Qfalse;
+    rb_ary_push(ret, boolean_val);
+    boolean_val = has_ref == TRUE ? Qtrue : Qfalse;
+    rb_ary_push(ret, boolean_val);
+
+    return ret;
+}
+
 void
 Init_glib_regex(void)
 {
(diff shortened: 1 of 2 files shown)


kou:
Sorry. I'm busy in this week. I'll check it later!

kou:
Thanks.
I've merged and split `GLib::Regex.check_replacement` to `GLib::Regex.check_replacement` and `GLib::Regex.have_references?` to implement Rubyish API.

kou:
> It remains to do the GLib::Regex#replace_eval which will be tricky. Do you want that I do the simple methods for GLib::MacthInfo before it?

You can do `GLib::Regex#replace_eval` before simple methods for `GLib::MatchInfo`.

> If you want that I do the GLib::Regex#replace_eval in the next step you should look here https://github.com/cedlemo/ruby-gnome2/blob/f38f53d2032e0df457d88a2c2d1a4f2f65b2d9e9/glib2/ext/glib2/rbglib_regex.c#L212 and https://github.com/cedlemo/ruby-gnome2/blob/f38f53d2032e0df457d88a2c2d1a4f2f65b2d9e9/glib2/ext/glib2/rbglib_regex.c#L254 . This is How I plan to implement it. Tell me if there is something wrong with this.

Could you extend `Glib::Regex#replace` instead of adding `GLib::Regex#replace_eval`?

I hope the following API:

```ruby
regex.replace("(pattern)", "[\\1]") # => "[pattern]"
regex.replace("pattern") do |text| # => "[pattern]"
  "[#{text}]"
end
```

cedlemo:
>I 've merged and split GLib::Regex.check_replacement to GLib::Regex.check_replacement and GLib::Regex.have_references? to implement Rubyish API.

Yes, I didn't feel that it was a good solution to return two values but I didn't take the initiative to transform this in something more rubyish.

> Could you extend Glib::Regex#replace instead of adding GLib::Regex#replace_eval?

Of course

cedlemo:
@kou,

For the implementation of the callback, I have the same feeling that I had for the implementation of `GLib::check_replacement`. Look here:

https://developer.gnome.org/glib/stable/glib-Perl-compatible-regular-expressions.html#GRegexEvalCallback

In my previous implementation I have done this:
https://github.com/cedlemo/ruby-gnome2/blob/f38f53d2032e0df457d88a2c2d1a4f2f65b2d9e9/glib2/ext/glib2/rbglib_regex.c#L212
```
/*  User can return in its callback:
   *  a string, the continue replacement is assumed
   *
   *  an array [string, Qnil] 
   *
   *  any others value are ignored
   *
   * */
  if (TYPE(returned_data) == T_STRING)
  {
    g_string_overwrite(result, 0, RVAL2CSTR(returned_data));
  }
  else
  {  
    if (TYPE(returned_data) == T_ARRAY)
    {
      VALUE string = rb_ary_entry(returned_data, 0);
      if(TYPE(string) == T_STRING)
        g_string_overwrite(result, 0, RVAL2CSTR(string));
      
      if (rb_ary_entry(returned_data, 1) == Qfalse)
        stop_replacement = FALSE;
    }
  }
  
  return stop_replacement;
```

As you can see, the user can use in the block a string or an array with a string and a value for continuing or stopping the replacement.

I am not sure that it will be OK for you. Maybe you can see a better solution?

kou:
The API is OK for now.
I'll reconsider about the API after you implemented.

PR #713, opened 2016-02-12
Add new example icons-theme-viewer.rb

cedlemo:
This example illustrate usage of `Gtk::IconTheme`, `Gtk::IconView` and `Gtk::Entry` and especially the secondary icon usage for the later.


kou:
Thanks!

PR #715, opened 2016-02-18
Add menus_from_resources.rb script and its UI files

cedlemo:
Add example that shows 3 types of menus and actions generated form UI files and gresources


kou:
Thanks!

PR #717, opened 2016-02-21
Implement g_regex_replace_eval

cedlemo:


cedlemo:
rebase done

kou:
Sorry for my late response...
I'm considering better API based on your work. I'll check it again later...

kou:
Thanks.
I've improve API based on your work.

cedlemo:
I have seen, it and I find it very instructive.

If I well understand in the callback , you have used parts of the code in ruby that handled the break statement in order to link the break statement to the "don't continue replacement" action.

I have just one little question do you recommend the systematic usage of `rb_protect` when one use `rb_funcall` ?

Thanks

I will send you now some pull-request for `MatchInfo`.

kou:
> If I well understand in the callback , you have used parts of the code in ruby that handled the break statement in order to link the break statement to the "don't continue replacement" action.

Right.

> I have just one little question do you recommend the systematic usage of `rb_protect` when one use `rb_funcall` ?

`rb_protect()` should be used when we don't handle code jump caused by `break`, exception and so on. Executing Ruby code in C level callback is the case. 

For example, we should use `rb_protect()` in C level callback. If code jump is caused in C level callback. C level cleanup code isn't executed.

```c
static void
some_callback(void)
{
  rb_raise(...);
  /* here is never executed. */
}

static void
some_function(void)
{
  some_callback();
  /* here is also never executed. */
  some_cleanup();
}
``` 

The `rb_protect()` usage for this case (getting `break` value) is special case. We'll not use the usage in other code.


> I will send you now some pull-request for `MatchInfo`.

Thanks!

PR #718, opened 2016-02-23
Gtk3 sample gtk demo printing rb

cedlemo:


gtk3/sample/gtk-demo/printing.rb
```diff
@@ -1,80 +1,40 @@
...
 
-    def on_begin_print(operation, context, data)
+    print_operation.signal_connect "begin-print" do |operation, context|
       height = context.height - HEADER_HEIGHT - HEADER_GAP
-      data.lines_per_page = (height / data.font_size).floor
-      data.lines = File.readlines(__FILE__)
-      data.n_pages = (data.lines.size - 1) / data.lines_per_page + 1
-      operation.set_n_pages(data.n_pages)
+      lines_per_page = (height / font_size).floor
+      file = File.open(resource_name, "r")
```
kou: Use

```ruby
File.open(...) do |file|
  # ...
end
```

or

```ruby
File.readlines(...)
```

The current code doesn't close an opened file.

cedlemo:
> The current code doesn't close an opened file.

Yes and variable name issues.

kou:
Thanks!

PR #719, opened 2016-02-23
Link the run button to an action

cedlemo:
Enable the possibility to run the currently selected demo in the treeview via the run button


kou:
Thanks!

PR #725, opened 2016-03-10
Glib2 match info partial match

cedlemo:
Here is the how I visualize what I will do next:

*    `GLib::MatchInfo#expand_references` for `g_match_info_expand_references`

For the different fetch functions, I first thought to make just one method that would have an hash parameter like we did previously for `GLib::Regex#match` for exeample, but I didn't think that it would be a good idea.
 
*    `GLib::MatchInfo#fetch` for `g_match_info_fetch`
*    `GLib::MatchInfo#fetch_named` for `g_match_info_fetch_named`
*    `GLib::MatchInfo#fetch_pos` for `g_match_info_fetch_pos`
*    `GLib::MatchInfo#fetch_named_pos` for `g_match_info_fetch_named_pos`
*    `GLib::MatchInfo#next` for `g_match_info_next`
*    `GLib::MatchInfo#fetch_all` for `g_match_info_fetch_all`

Tell me if you see a better way to do this.

### DIFF
--- a/glib2/ext/glib2/rbglib_matchinfo.c
+++ b/glib2/ext/glib2/rbglib_matchinfo.c
@@ -48,6 +48,12 @@ rg_match_count(VALUE self)
     return INT2NUM(g_match_info_get_match_count(_SELF(self)));
 }
 
+static VALUE
+rg_partial_match_p(VALUE self)
+{
+    return CBOOL2RVAL(g_match_info_is_partial_match(_SELF(self)));
+}
+
 void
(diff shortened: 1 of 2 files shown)


kou:
Please bind `g_match_info_fetch()` and `g_match_info_fetch_named()` to `GLib::MatchInfo#fetch` first.

If the argument is `Integer`, we use `g_match_info_fetch()`.
If the argument is `String`, we use `g_match_info_named()`.
Otherwise, we raise `ArgumentError`.

PR #731, opened 2016-03-19
Gtk3 sample gtk demo font features rb

cedlemo:


gtk3/sample/gtk-demo/font_features.rb
```diff
@@ -0,0 +1,113 @@
...
+      case name
+      when "reset"
+        proc do |button|
+          numcasedefault.active = true
+          numspacedefault.active = true
+          fractiondefault.active = true
+          toggle.each do |widget|
+            if widget.class != Gtk::RadioButton
+              widget.active = false
+              widget.set_property("sensitive", false)
```
kou: Please use method call style:

```ruby
widget.sensitive = false
```

gtk3/sample/gtk-demo/font_features.rb
```diff
@@ -0,0 +1,113 @@
+# Copyright (c) 2015 Ruby-GNOME2 Project Team
```
kou: 2016!

gtk3/sample/gtk-demo/font_features.rb
```diff
@@ -0,0 +1,113 @@
...
+        if widget.active?
+          s += ", " if has_feature
+          s += widget.builder_name
+          s += " 1"
+          has_feature = true
+        end
+      else
+        s += ", " if has_feature
+        s += widget.builder_name
+        widget.active? ? s += " 1" : s += " 0"
```
kou: `if` is better for expression that doesn't use return value:

```ruby
if widget.active?
  s += " 1"
else
  s += " 0"
end
```

gtk3/sample/gtk-demo/font_features.rb
cedlemo: Oups,  
it is because I have a script that gets the comments from the c source and create a skeleton with the copyright and launch neovim sessions in order and I have forgottent to update the date.

kou:
Thanks!

PR #737, opened 2016-03-30
Add News of the period after 2015-10-31 and before 2015-11-30

cedlemo:


kou:
Thanks!
There's no problem on your work. Could you try working on all range? (from 2015-12-01 to now)

PR #739, opened 2016-03-31
Finish the NEWS sumup

cedlemo:
I have added a global section because I didn't know where those changes should be added.

It takes me a lot of time. Maybe we should do this more frequently?

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -268,7 +268,7 @@ It improves GTK+ 3 support release. Or something...
   * Yuuki Harano
   * detlef
   * Aurélien Jacobs
-  * Hirochi Hatake
+  * Hiroshi Hatake
   * Malo Skrylevo
 
 == Ruby-GNOME2 3.0.7: 2015-10-06
--- a/NEWS
+++ b/NEWS
@@ -8,6 +8,14 @@ It improves GTK+ 3 support release. Or something...
 
 === Changes
 
+==== Global
+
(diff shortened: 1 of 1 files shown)


NEWS
```diff
@@ -91,7 +261,15 @@ It improves GTK+ 3 support release. Or something...
...
   * Christopher L. Ramsey
-  * ...
+  * Hiroyuki ITo
+  * dutchhome Jon Raiford
+  * Rob Brackett
+  * Abby Archer
+  * Yuuki Harano
+  * detlef
+  * Aurélien Jacobs
+  * Hirochi Hatake
```
cosmo0920: It's sad for my name typo.... :cry: 

```diff
-Hirochi
+Hiroshi
```

cedlemo: Oh I really embarassed for this! 
ごめんなさい!! (I hope this is the good way to say that I am sorry)

cosmo0920: Don't worry, we just fix it when someone find wrong things! :relaxed:

kou:
Thanks!!!

cosmo0920:
:blush:

kou:
> I have added a global section because I didn't know where those changes should be added.

We're using "All" section for it. I changed the section name.

kou:
> It takes me a lot of time. Maybe we should do this more frequently?

If it reduces resources for a new release, I agree with it.

Issue #736, opened 2016-03-28
Legacy code in the samples


ghost:
Hello Kouhei Sutou (and others),

I have made a small edit lately; as I am still new to git/github, I don't want to mess anything up.

I believe this line of code is no longer necessary on modern ruby:

```
require 'thread'
```

I have removed it from this file here:

https://github.com/ruby-gnome2/ruby-gnome2/blob/master/gtk2/sample/misc/threads.rb

And I have tested locally, it still works when I invoke it via "ruby threads.rb" so I believe that this is legacy code. I can go through the other sample files and try to make small modifications, but before I do so, I would like to hear confirmation. Thank you. :)

I can try to go through some of the older examples systematically too and correct small problems. I am not yet too familiar with gnome3/gtk3 so I am not sure how much I can do for the latter.

(This issue can be closed, it just was my aim for trying to get some feedback before I proceed. Thanks.)

kou:
Thanks for your work!
Could you create a pull request? We can check before merging your change to master.

shevegen:
Ok, give me a bit to find out how to do that precisely. :D

cedlemo:
@shevegen,

I am working for a moment now on the samples in gtk3. We have already updated the scripts in `gtk3/samples/misc`, adapted the official the gtk+ tutorial in `gtk3/samples/tutorial` and It remained the `gtk3/samples/gtk-demo`. 

We have started this part, the main.rb was created and we have almost updated half of the demos. Now the gtk+ 3.20 has been released and the main.rb must be updated, some demos that have done must be updated too and it remains, of course, the demos that have not been done.

Do you want to help me ? I will provide you more explications on what I have done, how and what must be done and we will see together how we will organize our work if you are OK. You said : 
> I am not yet too familiar with gnome3/gtk3
I wasn't when I started (I am still far from an expert) but I started the update in order to learn how it works.

Regards

PR #747, opened 2016-04-10
Gtk3 sample gtk demo 3 20 changes

cedlemo:
Some demos will need to be adapted to the new change and the main.rb too. Some works in sight for me !! :-)

Do you think that shevegen could be interested in helping me #736


kou:
Thanks!!

> Do you think that shevegen could be interested in helping me #736

I think that @shevegen will be interested in it. :-)

PR #757, opened 2016-04-27
Finish update 3.20 for the demos in 3.19 version

cedlemo:
All the demos that were updated in the gtk 3.19 version are now updated to the 3.20. if needed.


kou:
Thanks!!!

PR #760, opened 2016-05-01
Fix error with MACRO usage

cedlemo:
I was looking how you handled the callbacks in pango and I have found this macro inversion.


kou:
Good catch!

PR #761, opened 2016-05-04
Pango ext pango cairo set shape renderer

cedlemo:
I have created a test file for the `Pango::Context` class and I have added tests that were missing.
(I thought that we always do a test for each methods that was implemented.) 

I was not able to implement some tests (you can find them with the TODO keyword). 

*   `test_load_font` and `test_load_fontset` don't work
*   `test_get_metrics` and `test_famillies` are not implemented.
*   You will have to look to the test `test_set_shape_renderer` that don't work neither. For example the line `assert_instance_of(Pango::Font, attr)` didn't generate an error.

I try to help you but I feel like I give you some more work. :-)


kou:
Thanks.
Could you split changes that add tests for existing methods to another pull request? I want to work on them as a separated task.

cedlemo:
I have remove the tests that where not related to the `set_shape_renderer` method and created another pull request with the other tests #763 . I hope it fits your need.

kou:
Thanks.
We'll handle #763 before this pull request.

cedlemo:
Ok, No Pb, I will do a rebase on this one after that.

kou:
Thanks.
Could you move test to `gtk3/test-pango-context.rb`? We should put tests that depend on gtk3 into `gtk3/test/` like `gtk3/test-gdk-event.rb`.

cedlemo:
I prefered to create a new PR #764 .

PR #764, opened 2016-05-08
Pango ext pango context set shape renderer

cedlemo:
Sorry for the delay,

I have created a new pull request. It is cleaner.


kou:
Thanks!!

PR #765, opened 2016-05-11
Gtk3 sample misc bindings rb update css

cedlemo:
Detlef reported having issue with the Css that didn't work recently in ruby-gnome2-devel-en.

This file is the only one that needed to be updated in all the demo/samples.

I need to check the gtk tutorial. I hope be able to do it in the next days.


kou:
Thanks!

PR #767, opened 2016-05-12
Gtk3 sample gtk demo rotated text rb

cedlemo:


kou:
Thanks.
By the way, remaining `rotated_text_old.rb` is OK?

cedlemo:
No remaining `rotated_text_old.rb` in not OK at all. 
I have been careless and it should be removed.

PR #768, opened 2016-05-13
Remove rotated_text_old.rb

cedlemo:


kou:
Thanks for confirming it.

PR #771, opened 2016-05-18
Gtk3 sample gtk demo pixbufs rb

cedlemo:
If you look at the line 119, in the C version they use `gdk_pixbuf_new_from_resource` I was not able to create directly a `Gdk::Pixbuf` from a resource so I created a `Gtk::Image` and use the `Gyk::Image#pixbuf` method.

Tell me if think that we should use `Gdk::Pixbuf.new(:resource => path)`


kou:
We should implement `Gdk::Pixbuf.new(:resource => path)`!

PR #778, opened 2016-05-22
Append the gsf, gnm and goffice in the Rakefile

cedlemo:
I was not able to build the gem files with `rake gem:build` so I modified the main Rakefile. I hope this is the good way to do this.


kou:
Thanks!!

PR #781, opened 2016-05-27
Gi loader rubyish class name

cedlemo:
I have used the method `rubyish_class_name` in

*    load_object_info
*    load_enum_info
*    load_struct_info
*    define_struct
*    load_flags_info
*    load_interface_info

and  I have added a comment af line 177. There is this function:

```ruby
    def define_enum(info)
      self.class.define_class(info.gtype,
                              enum_class_name(info),
                              @base_module)
    end

    def enum_class_name(info) # TODO : should be removed? used only in Gtk3 loader
      info.name
    end
```
The `enum_class_name` is only used in Gtk3, I think that we should replace it with `rubyish_class_name`

### DIFF
--- a/gobject-introspection/lib/gobject-introspection/loader.rb
+++ b/gobject-introspection/lib/gobject-introspection/loader.rb
@@ -121,11 +121,12 @@ module GObjectIntrospection
     end
 
     def define_struct(info, options={})
+      name = rubyish_class_name(info.name)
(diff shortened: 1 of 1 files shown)


cedlemo:
@kou in case you didn't see this PR

kou:
Thanks for the remainder!

PR #787, opened 2016-06-02
Gdk pixbuf2 gi

cedlemo:
@kou ,

I started to implement the `Gdk::Pixbuf` with GObjectIntrospection. I push this because I had to do some choice and I want your opinion before going further.

*   the main idea is to create a GdkPixbuf module, fill it with a `GObjectIntrospection::Loader` and after that include it in the `Gdk` module.

*   For the `GdkPixbuf::Pixbuf.new`, there are a lot of `initialize_from*` so I decided to use an hash as argument ( in gdk_pixbuf2/lib/gdk_pixbuf2/pixbuf.rb ).
In the previous code, the user could create a pixbuf like this:

```ruby
Gdk::Pixbuf.new(filename)
Gdk::Pixbuf.new(colorspace, has_alpha, width, size ...)
# or with hash form
```
I think that it is very confusing and that we should only use the hash form (like we do in Gtk3 for example)

*   You told me to make Gdk::Pixbuf work with all the samples, I think that it will be very difficult because  some of the code are deprecated. I think that I should focus on doing good tests in `gdk_pixbuf2/test` or `gtk3/test` if needed.

All of this are suggestions. I wait for your advices before going further.

Regards


kou:
Thanks.

> the main idea is to create a GdkPixbuf module, fill it with a GObjectIntrospection::Loader and after that include it in the Gdk module.

It's OK.

> For the GdkPixbuf::Pixbuf.new, there are a lot of initialize_from* so I decided to use an hash as argument ( in gdk_pixbuf2/lib/gdk_pixbuf2/pixbuf.rb ).

It's OK.

> In the previous code, the user could create a pixbuf like this:
> ...
> I think that it is very confusing and that we should only use the hash form (like we do in Gtk3 for example)

It's OK but we should also provide the previous APIs for backward compatibility. It's acceptable that we warn against the previous APIs.

> You told me to make Gdk::Pixbuf work with all the samples, I think that it will be very difficult because some of the code are deprecated. I think that I should focus on doing good tests in gdk_pixbuf2/test or gtk3/test if needed.

Does it mean that we drop `gdk_pixbuf2/sample/*.rb`?
It's good that we create tests in `gdk_pixbuf2/test` and/or `gtk3/test`. Because we can confirm that gdk_pixbuf2 isn't broken with them.
`gdk_pixbuf2/sample/*.rb` are used for confirming backward compatibility. They aren't for confirming whether gdk_pixbuf2 is broken or not.

PR #788, opened 2016-06-02
Gdk pixbuf2 backward compatibility

cedlemo:
I hope this is what you expected. Ifollowed the C code in rbgdk-pixbuf.c. All the tests are not done yet but I prefer to be sure that my work correspond to what you asked me.

One question about the inclusion of Gdk_pixbuf2, the only way to make it works was to do something like this : 

```ruby
module Gdk
    class Pixbuf < GdkPixbuf::Pixbuf
    end
end
```
Is there any "metprogramming" tricks that can try to load from  GdkPixbuf all the non found constants or methods of Gdk like with `method_missing`.

### DIFF
--- a/gdk_pixbuf2/lib/gdk_pixbuf2/pixbuf.rb
+++ b/gdk_pixbuf2/lib/gdk_pixbuf2/pixbuf.rb
@@ -2,8 +2,57 @@ module GdkPixbuf
   class Pixbuf
     alias_method :initialize_raw, :initialize
 
-    def initialize(options = {})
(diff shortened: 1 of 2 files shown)


kou:
Thanks!

I've added codes for backward compatibility.

> Is there any "metprogramming" tricks that can try to load from GdkPixbuf all the non found constants or methods of Gdk like with method_missing.

We don't need to implement lazy initialization. Because gdk-pixbuf doesn't provide initialization function.

PR #789, opened 2016-06-02
Compile with glib2 <2.30.0

adamhooper:
It looks like ruby-gnome2 is meant to support glib2 < 2.30.0. But it doesn't compile, because of a one-line mistake. The fix seems simple enough.


kou:
Could you show error message?

cosmo0920:
CentOS 6 provides glib2 2.28.8: https://rpmfind.net/linux/RPM/centos/6.8/x86_64/Packages/glib2-2.28.8-5.el6.x86_64.html
But `G_TYPE_MATCH_INFO` is provided 2.30 or later: https://developer.gnome.org/gobject/stable/gobject-Boxed-Types.html#G-TYPE-MATCH-INFO:CAPS
I think this PR is reasonable fix.

kou:
Did you try it?

adamhooper:
I don't have time today to go reproduce the error message. But you can trust me on two facts:

1. the error message involved `G_TYPE_MATCH_INFO`
2. after my fix, the glib2 gem compiles and installs on CentOS 6 and Fedora 23 without error

kou:
I'll wait for error message from compiler.

I think that macro value isn't evaluated until the macro is used.

adamhooper:
Know what? I'll bet you're right. My compile error was on 3.0.8; I didn't test master before adding my tweak. It looks like the actual fix was in commit 460151eff10fcfd39a777b8e1abce282ef3901e7. I'll confirm tomorrow.

adamhooper:
Sorry for the long wait. Yup, it works without my change. The fix was already on master.

If you have the time, please consider releasing the fix. In the meantime I have to write the gem and all its dependencies to `vendor/cache/`, which is a bit of a maintenance mess.

PR #790, opened 2016-06-05
Include GdkPixbuf::Pixbuf into Gdk

cedlemo:
With this, all is working fine. What do you want me to do next?

Continue the gtk-demos or finish the tests for gdk_pixbufs. I would prefer finish the tests for the implemented options and functions of GdkPixbuf.

PR #783, opened 2016-05-31
Fix an error with ubuntu and its Gir files

cedlemo:
This is a try to fix the travis pb : 

On my system:

```
strings /usr/lib/girepository-1.0/GOffice-0.10.typelib | grep Component
Component
GOComponent
ComponentClass
ComponentMimeDialog
GOComponentMimeDialog
ComponentPrivate
ComponentSaxHandler
ComponentType
cComponent
GocComponent
```

On a vagran VM with trusty 64 (like Travis)
```
strings /usr/lib/girepository-1.0/GOffice-0.10.typelib | grep Component
Component
GOComponent
ComponentMimeDialog
GOComponentMimeDialog
ComponentPrivate
ComponentSaxHandler
ComponentType
_ComponentClass
cComponent
GocComponent
```

I think that the GIR or code doesn't match.

PR #791, opened 2016-06-07
Add tests for GdkPixbuf::Pixbuf#new

cedlemo:
@kou just a remainder that I have two pending questions: 

#783 : Is there a way to exclude some structure to be loaded in GI ?
#790 : I don't see how `GLib::Deprecatable` does the same job that the instruction include `GdkPixbuf`.

I don't know if you had seen them, just to be sure.

Thanks.

### DIFF
--- a/gdk_pixbuf2/test/test-pixbuf.rb
+++ b/gdk_pixbuf2/test/test-pixbuf.rb
@@ -65,6 +65,14 @@ class TestPixbuf < Test::Unit::TestCase
         pixbuf = GdkPixbuf::Pixbuf.new(r_xpm)
         assert_equal(GdkPixbuf::Colorspace::RGB, pixbuf.colorspace)
       end
+
+      test "from_file_at_size" do
+        pixbuf = GdkPixbuf::Pixbuf.new(fixture_path("gnome-logo-icon.png"),
+                                                    32, 48)
+        assert_equal(GdkPixbuf::Colorspace::RGB, pixbuf.colorspace)
+        assert_equal(32, pixbuf.width)
+        assert_equal(32, pixbuf.height)
+      end
     end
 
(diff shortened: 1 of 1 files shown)


gdk_pixbuf2/test/test-pixbuf.rb
```diff
@@ -65,6 +65,14 @@ def r_xpm
         pixbuf = GdkPixbuf::Pixbuf.new(r_xpm)
         assert_equal(GdkPixbuf::Colorspace::RGB, pixbuf.colorspace)
       end
+
+      test "from_file_at_size" do
+        pixbuf = GdkPixbuf::Pixbuf.new(fixture_path("gnome-logo-icon.png"),
+                                                    32, 48)
+        assert_equal(GdkPixbuf::Colorspace::RGB, pixbuf.colorspace)
+        assert_equal(32, pixbuf.width)
+        assert_equal(32, pixbuf.height)
```
kou: It should be `48`.

gdk_pixbuf2/test/test-pixbuf.rb
```diff
@@ -96,6 +104,16 @@ def r_xpm
...
       end
+
+      test "from_file_at_size" do
+        filename = fixture_path("gnome-logo-icon.png")
+        pixbuf = GdkPixbuf::Pixbuf.new(:file => filename,
+                                       :width => 32,
+                                       :height => 48)
+        assert_equal(GdkPixbuf::Colorspace::RGB, pixbuf.colorspace)
+        assert_equal(32, pixbuf.width)
+        assert_equal(32, pixbuf.height)
```
kou: Ditto.

kou:
Ah, sorry. The API keeps aspect ratio.

PR #794, opened 2016-06-07
Fix crashes in samples

Madh93:
Fixed crashes about:

**File:** `tooltips.rb`
**Problem:** Crash when mouse over GtkTextView element
**Cause:** `Gtk::TextView.get_iter_at_position()` returns [Boolean, Gtk::TextIter]

**File:** `textbuffer-serialize.rb`
**Problem:** Crash when set color/font to unselected region
**Cause:** `Gtk::TextBuffer.apply_tag()` unexpected nil bounds


kou:
Thanks for your report.
I've fixed them.

PR #795, opened 2016-06-09
The size ssting method was wrong

renich:
Was it a typo or did the method change?  I dunno. It works now.


kou:
Thanks. I've merged it.
Please don't include off-topic changes (such as space changes) at the next time.

renich:
got it. Thanks.

PR #801, opened 2016-06-16
Add test for GdkPixbuf::Pixbuf#fill!

cedlemo:


gdk_pixbuf2/test/test-pixbuf.rb
```diff
@@ -164,4 +164,11 @@ def test_new_subpixbuf
       assert_equal(32, pixbuf.height)
     end
   end
+  
+  def test_fill!
+    filename = fixture_path("gnome-logo-icon.png")
+    pixbuf = GdkPixbuf::Pixbuf.new(:file => filename)
```
kou: Could you create two `GdkPixbuf::Pixbuf` instances? Because `#fill` is destructive method.

cedlemo:
Done and rebase too.

kou:
Thanks!

PR #814, opened 2016-06-24
Use GdkPixbuf form in pixbufs.rb

cedlemo:
I have just updated this one demo with the `GdkPixbuf::Pixbuf` form because the others demos with `Gdk::Pixbuf` were yet not working.


kou:
Thanks!!
I've added `Pixbuf#composite!`.

PR #816, opened 2016-06-25
Segmentation fault in Poppler::Page#text_layout.

mephistobooks:
Hi.
I used poppler.gem 3.0.8 in Mac OS X 10.10.5, ruby version is 2.2.2 and 2.3.1. And I got segmentation fault when I access several times to the array of PopplerRectangle objects via Page#text_layout.

So I rewrote text_layout method in a simple way. Could you please see my modification and test case?
Thanks.


kou:
Thanks. It's a GC related problem.

You can reproduce it by the following code:

```ruby
document = Poppler::Document.new(form_pdf)
page = document[0]
GC.stress = true
layout = page.text_layout
GC.stress = false
GC.start
```

PR #824, opened 2016-06-27
Gdk pixbuf sample composite gdk pixbuf form

cedlemo:


kou:
Ah, we should provide Rubyish `#composite` API.
I'll consider about it.

kou:
I added options version. Backward compatibility is kept. (Ordered arguments version is also supported.)

PR #837, opened 2016-07-03
Start updating NEWS file

cedlemo:
From : 

    git log --decorate --stat --reverse --after="2016-04-03"

There are some parts that you should look at (at the very beginning).


kou:
Thanks!

PR #838, opened 2016-07-05
Add tests for GdkPixbuf::Pixbuf#saturate_and_pixelate

cedlemo:


kou:
Patch by cedlemo. Thanks!!!

Issue #782, opened 2016-05-29
Remove ifdef HAVE_RUBY_ENCODING_H


kou:
We already dropped Rubies that doesn't support encoding.

kou:
Done.

PR #840, opened 2016-07-05
Remove conditionnal macro based on HAVE_RUBY_ENCODING_H

cedlemo:
If I understand well the ruby/encoding.h has been introduced with ruby 1.9 ~ so you want to remove the if close because you support only ruby version (> 2.1).

I hope it is what you wanted in #782.


kou:
Thanks!

Issue #839, opened 2016-07-05
`GdkPixbuf::Pixbuf#formats` crash


cedlemo:
@kou,

When I try to list the formats that `GdkPixbuf::Pixbuf` supports in order to save a pixbuf, it crashes.

Minimum code:

```Ruby
require "gtk3"

puts GdkPixbuf::Pixbuf.formats
```

```
ruby gdk_pixbuf_saving_formats.rb 
/home/cedlemo/.gem/ruby/2.3.0/gems/gobject-introspection-3.0.9/lib/gobject-introspection/loader.rb:114: [BUG] Segmentation fault at 0x00000000000000
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-linux]

-- Control frame information -----------------------------------------------
c:0004 p:---- s:0013 e:000012 CFUNC  :invoke
c:0003 p:0065 s:0009 e:000008 LAMBDA /home/cedlemo/.gem/ruby/2.3.0/gems/gobject-introspection-3.0.9/lib/gobject-introspection/loader.rb:114 [FINISH]
c:0002 p:0024 s:0005 E:002390 EVAL   gdk_pixbuf_saving_formats.rb:3 [FINISH]
c:0001 p:0000 s:0002 E:000be0 (none) [FINISH]

-- Ruby level backtrace information ----------------------------------------
gdk_pixbuf_saving_formats.rb:3:in `<main>'
/home/cedlemo/.gem/ruby/2.3.0/gems/gobject-introspection-3.0.9/lib/gobject-introspection/loader.rb:114:in `block in define_singleton_method'
/home/cedlemo/.gem/ruby/2.3.0/gems/gobject-introspection-3.0.9/lib/gobject-introspection/loader.rb:114:in `invoke'

-- Machine register context ------------------------------------------------
 RIP: 0x00007f5275557702 RBP: 0x0000000002bf0b20 RSP: 0x00007ffc6ed84b28
 RAX: 0x0000000000000000 RBX: 0x61647869506b6447 RCX: 0x00000000026457b0
 RDX: 0x0000000000000000 RDI: 0x61647869506b6444 RSI: 0x61647869506b6447
  R8: 0x000000000000000c  R9: 0x0000000001eb99e0 R10: 0x00000000000003af
 R11: 0x00007f527579d6b0 R12: 0x00007ffc6ed84b30 R13: 0x61647869506b6447
[...]

PR #842, opened 2016-07-09
Support gslist of struture values as return value

cedlemo:
For the crash when `GdkPixbuf::Pixbuf#formats` is called #839


kou:
Thanks!

Issue #845, opened 2016-07-26
Support rb_cInteger in ruby-head


cosmo0920:
In Current Ruby HEAD (2.4.0),  `rb_cFixnum(Fixnum)` and `rb_cBignum(Bignum)` seems to be unified in `rb_cInteger(Integer)`.
This causes build error with ruby-head:
https://travis-ci.org/ruby-gnome2/ruby-gnome2/jobs/147253146#L1517
Should we support `rb_cInteger(Integer)` in glib2 for now?

If we decide to support `rb_cInteger` and keep backward compatibility in glib2, `RUBY_INTEGER_UNIFICATION` macro is helpful for us.
ref: https://svn.ruby-lang.org/cgi-bin/viewvc.cgi?view=revision&revision=55394

kou:
Done.

PR #848, opened 2016-07-29
glib2: support Ruby 2.4 Integer unification

cosmo0920:
Related to #845.

ref: https://bugs.ruby-lang.org/issues/12005


kou:
Thanks!

PR #854, opened 2016-08-08
Update NEWS file

cedlemo:


kou:
Thanks!
Can we remove TODO in NEWS?

cedlemo:
Of course, I let you there in case you had something to add that I have forgotten.
by the way sorry for the wait I couldn't do this update earlier.

kou:
Thanks. I've removed it.

> by the way sorry for the wait I couldn't do this update earlier.

No problem. :-)

Issue #859, opened 2016-08-14
Problem getting gdk3/ to run and work


shevegen:
Hello.

The README.md at the toplevel states this:

    % cd <each sub-directory>
    % ruby extconf.rb
    % make
    % sudo make install

In the subdirectory called:

  gtk3/

This works fine. There is a file called extconf.rb

In the subdirectory called:

  gdk3/

There is no extconf.rb file. There is a Rakefile
but I can not get it to run.

Would it be possible to add a extconf.rb file there or
perhaps unify the install instructions?

For some odd reason, extconf.rb always works fine for
me but Rakefiles I often have a lot of more problems
with.

The reason why I mention this is because I am trying
to get the sample code in gtk3/ to work, but with gdk3
having a problem, I can not get it to work so I am stuck
with gtk2 for now (which works fine).

I somehow got it to work or so I think but I have a second
problem, a crash:

  ruby rubygnome2-3.0.9/gtk3/sample/gtk-demo/main.rb 

error is:

/Programs/Ruby/2.3.1/lib/ruby/site_ruby/2.3.0/gobject-introspection/loader.rb: line 564                                                                                                       
   Gtk-DEBUG **:Connecting to session manager                                                                                                                                                 
/Programs/Ruby/2.3.1/lib/ruby/site_ruby/2.3.0/glib2/deprecatable.rb:112:in `const_missing': uninitialized constant Gtk::TextBuffer::GdkPixbuf   

I think I may have some gdk_pixbuf problem on my system - it
would be nice if some error message could be used to indicate
what is going wrong where, if this is possible.

Thanks for reading!

kou:
Please use `gem install gtk3` instead of `extconf.rb`. It's only for developers not users.

kou:
We need to update README.md.

PR #860, opened 2016-08-14
Update README

cedlemo:
I have moved the installation instruction to the begining. I have added an icon of the ruby-GNOME2 project and an image that displays the current version of gtk3 (which is the version of all the gems).

Feel free to close this PR if you don't like it.

#859


kou:
Thanks! I like it.

PR #894, opened 2016-08-26
Gtk3 sample gtk demo drawingarea rb

cedlemo:
@kou, could you check this demo at lines 165-170. I can not find which way is the better in order to define the events we want to catch.

It looks like in the demo, the da is not able to catch motion events.

Thx.


kou:
> could you check this demo at lines 165-170. I can not find which way is the better in order to define the events we want to catch.

If we want to add events, `#add_events` is preferred. If we want to modify events, `events=` is preffered.

> It looks like in the demo, the da is not able to catch motion events.

da catch motion events but there is a bug in motion event handling. I've fixed it.

PR #906, opened 2016-09-05
gdk_pixbuf2: Fix runtime dependency

cosmo0920:
gdk_pixbuf2 also requires 'gobject-introspection' gem.

Follows up https://github.com/ruby-gnome2/ruby-gnome2/commit/5cb0def4dad4439b9e4aea431578dfc7b98a2883.
I've also encountered to the same issue: https://github.com/ruby-gnome2/ruby-gnome2/issues/900#issuecomment-244742030.

### DIFF
--- a/gdk_pixbuf2/Rakefile
+++ b/gdk_pixbuf2/Rakefile
@@ -23,6 +23,7 @@ package_task = GNOME2::Rake::PackageTask.new do |package|
   package.summary = "Ruby/GdkPixbuf2 is a Ruby binding of GdkPixbuf-2.x."
   package.description = "Ruby/GdkPixbuf2 is a Ruby binding of GdkPixbuf-2.x."
   package.dependency.gem.runtime = [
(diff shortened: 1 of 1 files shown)


cosmo0920:
Now, `gdk_pixbuf2` had been used gi based binding. We should require `gobject-introspection` at runtime.

Kagetsuki:
+1, but are we sure we actually need gobject-introspection? The dependency ends up requiring a much more involved native build chain and many more dependencies, but I'm not sure it's even doing anything in gdk_pixbuf2. It seems to have just been added but I can't seem to find the reason why.

cosmo0920:
> but are we sure we actually need gobject-introspection? 

Yes, gdk_pixbuf2 has been completely rewritten by gobject-introspection gem based Ruby code.

But, sorry. It's meaningless PR. 
Because gio2 already requires `gobject-introspection`.

I'm closing this PR for now.

PR #910, opened 2016-09-05
do not try to run a demo when the user provide a bad demo name

cedlemo:
this prevent to try to run a demo when a user do `main.rb -r a_bad_demo_name`.


kou:
Thanks!!!

PR #911, opened 2016-09-05
Gtk3 sample gtk demo main rb set focus on treeview

cedlemo:
When the user use : 

    ./main.rb -r revealer

Then the related item in the treeView is selected and the user can see the information tab and source tab for the *revealer* demo.


kou:
Could you rebase on master? It's conflicted.

cedlemo:
Done.

kou:
Thanks!

PR #934, opened 2016-09-30
Gtk3 sample gtk demo listbox rb

cedlemo:
I would like to point out that the demo takes some times to load. And that it doesn't work exactly like the C demo, the extra_buttons_box in each row is shown even if the row is not selected when the demo starts.


kou:
Thanks.
I also confirmed them. They should be fixed later...

Issue #943, opened 2016-10-10
Release 3.1.0


kou:
TODO:
- [x] Update NEWS
- [x] #898
- [x] #897
- [x] #916
- [x] #928
- [x] #935

kou:
The remaining task is only "Update NEWS".
@cedlemo, could you work on it?

cedlemo:
Of course.

kou:
Now, the rest task is releasing gems.

kou:
I've released 3.1.0! (Fat gems for Windows are building now.)
@cedlemo, could you send release announce?

kou:
I've pushed fat gems for Windows.

cedlemo:
> could you send release announce? 

Of course.

cedlemo:
Done

kou:
Thanks!

PR #945, opened 2016-10-12
Clutter gdk

cedlemo:
#943 

I was able to successfully load this lib by adding this line in the `clutter-gtk.rb` file : 

          ClutterGdk.init if ClutterGdk.respond_to?(:init)

Then the following test was working : 

```ruby
require "gtk3"
require "clutter-gtk"
ClutterGtk.init
win = Gtk::Window.new(:toplevel)

win.signal_connect("destroy") { Gtk.main_quit }

but = Gtk::Button.new(:label => "click")
but.signal_connect "clicked" do |widget|
 stage = ClutterGdk.get_stage_from_window(win.window)
 puts stage
end
win.add(but)
win.show_all
win.set_default_size(400, 400)
Gtk.main
```

The only problem I saw was that this line `stage = ClutterGdk.get_stage_from_window(win.window)` returns `nil`. In the same way this  : 

```ruby
puts ClutterGdk.default_display.class #=>  #<Class:0x00000003f54d50>
```
returns an unknown class.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -118,6 +118,7 @@ gnome2_packages = gtk2_packages + gtk3_packages + [
   "vte3",
   "gtksourceview3",
   "clutter",
+  "clutter-gdk",
   "clutter-gtk",
   "clutter-gstreamer",
   "webkit-gtk",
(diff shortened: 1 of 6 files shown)


kou:
Thanks!

PR #957, opened 2016-11-12
Update NEWS from 2016-08-12 to 2016-11-12

cedlemo:
Here is the NEWS file updated. The last commit I used is 55b24df4bbe128f55cda0997f91e7903a9c06f98.


cedlemo:
@kou in case you missed it.

kou:
Thanks! Very fast!

PR #958, opened 2016-11-13
Add option hash as possible argument to GtkTreeViewColumn::new and add tests for it

detlefw:


cedlemo:
ping @kou :)

kou:
Thanks!
Sorry. I missed this.

PR #961, opened 2016-11-15
Glib2 datetime class

cedlemo:


kou:
Thanks! Sorry for late!

cedlemo:
No problem.

kou:
I'll change `GLib::DateTime.now_local` to `GLib::DateTime.now(:local)`. So please wait a bit to use it.

kou:
Done.

PR #965, opened 2016-11-28
Glib2 datetime new

cedlemo:


kou:
Thanks.

I changed some API:

Before:

    :unix_local => ...

After:

    :unix => ...,
    :timezone => :local

Before:

    :unix_utc => ...

After:

    :unix => ...,
    :timezone => :utc

Before:

    :offset => "local", ...

After:

    :timezone => :local, ...

Before:

    :offset => "utc", ...

After:

    :timezone => :utc, ...

cedlemo:
> I changed some API

I find yours APIs better than mine.

PR #966, opened 2016-11-29
Gtk3 sample gtk demo old demos

cedlemo:
This will prevent users to try to run old demos, that I want to keep, via the `main.rb`.


kou:
I agree with this.

PR #973, opened 2016-12-12
gtk-demo main.rb : when demo is launched from command line and belong…

cedlemo:
… to child in the treeview, expand before select the row


kou:
Thanks!!

PR #975, opened 2016-12-25
Gio2 lib settings initialize

cedlemo:
@kou,

could you check the tests, I tried to stay close to the previous organization of the files you created, but IMO with my modifications, it looks like it is a mess now.

thanks.


kou:
Thanks.
I clean up gschema related code.

BTW, does `Gio.keyfile_settings_backend_new` exist on your environment? It doesn't exist on my Debian GNU/Linux sid environment.

cedlemo:
Yes it does, all the code I pushed works in my environment. I am using Archlinux, so maybe my gio libs are newer.

cedlemo:
here : 
https://developer.gnome.org/gio/2.50/GSettingsBackend.html

It says since `gio 2.28`

In my env : 

    > Gio::Version::STRING
    => "2.50.2"

kou:
Thanks.
It seems that `Gtio.keyfile_settings_backend_new` exists since GObject Introspection 1.51.0. (`Gio-2.0.gir` and `Gio-2.0.typelib` are provided by GObject Introspection not GLib.)
I'm using 1.50.0.

cedlemo:
@kou,

I don't understand, I thought that the availability of a method was in fact dependant to their existence in the lib and totally independent of the GObject instrospection version.

did I misunderstood ?

kou:
GObject Introspection extract method information from document in source.
Gio's source includes `gio_keyfile_settings_backend_new()` entry but GObject Introspection disabled it until 1.51.0: https://git.gnome.org/browse/gobject-introspection/commit/?id=b8d92b0b36b3907ef066e068e33e9309eb0f8ec5

So GObject Introspection 1.50.0 provides `Gio-2.0.gir` that doesn't include `gio_keyfile_settings_backend_enw()`.

cedlemo:
thanks!

kou:
No problem!

PR #977, opened 2016-12-28
Extconf

detlefw:
If ruby gnome is build with make after calling extconf.rb in the projects root, gdk3 and gdk_pixbuf2 will not get build, because the do not have an extconf.rb file in the sub-projects directory. This commit ads them.

### DIFF
--- a/gdk3/extconf.rb
+++ b/gdk3/extconf.rb
@@ -1,6 +1,6 @@
 #!/usr/bin/env ruby
 #
-# Copyright (C) 2015  Ruby-GNOME2 Project Team
+# Copyright (C) 2015, 2016  Ruby-GNOME2 Project Team
 #
 # This library is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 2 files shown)


kou:
We don't need them because we don't need to build C sources for gdk3 and gdk_pixbuf2.

detlefw:
But without it the Makefile.lib for gdk3 and gdk_pixbuf2 are not 
generated and so they will not get installed, if you want to install 
ruby-gnome by doing a extconf.rb && make && make install in the top 
level directory.

Am 29.12.2016 um 14:30 schrieb Kouhei Sutou:

kou:
Use `rake gem:build gem:install` instead.

detlefw:
I want to avid gems as far as possible. One of the reasons is, that I 
have to install the gems with every new version of ruby and that for all 
users on a system. For that extconf is very handy and I prefer, that it 
would stay usable.


Am 29.12.2016 um 14:42 schrieb Kouhei Sutou:

kou:
Use `gem install --user-install ...`. It installs a gem only for you not for all users on a system.

PR #979, opened 2016-12-28
Ignore build time-stamps.

detlefw:
Ignore build-time stamps for git, which are created by a extconf.rb / make run.


kou:
Thanks!!

FYI: We recommend to use `rake gem:build gem:install` instead of `make install`.

PR #984, opened 2017-01-07
Fix gtksourceview3 sample

mvz:
The #add method does not return self, so an intermediate variable is needed.


kou:
Thanks. But we never maintain no-gi versions.

And I changed `Gtk::Container#add` returns `self`. Because we should keep backward compatibility as much as possible.

mvz:
Maybe this example can be moved to the gi version then?

kou:
Yes!

PR #1016, opened 2017-04-19
Improve tests for GObjectIntrospection::UnionInfo

cedlemo:
I don't know any union with `#is_discriminated? == true`. If you know one please tell me, so that I can test all the functions of `GObjectIntrospection::UnionInfo`.


kou:
Thanks!
I also don't know...

PR #1019, opened 2017-04-25
Fix check for inheritance from GLib::Instantiatable

mvz:
This fixes the check that Interface modules do so they can only be included in classes that inherit from `GLib::Instantiatable`.


kou:
Thanks!

PR #1028, opened 2017-05-06
Make WebView constructors available

mvz:
This makes all `new_with_<something>` constructors available for WebView, similar to Pixbuf.

### DIFF
--- a/webkit2-gtk/lib/webkit2-gtk.rb
+++ b/webkit2-gtk/lib/webkit2-gtk.rb
@@ -67,6 +67,7 @@ module WebKit2Gtk
 
     def require_libraries
       require "webkit2-gtk/version" if @version_module.const_defined?(:MAJOR)
+      require "webkit2-gtk/web_view"
     end
 
     def load_constant_info(info)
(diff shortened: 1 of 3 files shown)


kou:
Thanks.
I've merged and changed your codes.

It seems that `#new_with_related_view` is a WebKit2Gtk bug. I think that `webkit_web_view_new_with_related_view()` should be treated as a constructor but it's treated as a method. Because the `webkit_web_view_new_with_related_view()`'s first argument type is `WebKitWebView *`. I don't confirm but the following patch is needed:

```diff
Index: WebKitWebView.cpp
===================================================================
--- WebKitWebView.cpp	(revision 216326)
+++ WebKitWebView.cpp	(working copy)
@@ -2225,7 +2225,7 @@
 }
 
 /**
- * webkit_web_view_new_with_related_view:
+ * webkit_web_view_new_with_related_view: (constructor)
  * @web_view: the related #WebKitWebView
  *
  * Creates a new #WebKitWebView sharing the same web process with @web_view.
```

See also: https://wiki.gnome.org/Projects/GObjectIntrospection/Annotations#Support_for_GObject_objects

Could you report it to WebKit2Gtk developers?

mvz:
> Could you report it to WebKit2Gtk developers?

(Finally) done!

kou:
Great!

PR #1030, opened 2017-05-12
Rsvg2 gi

cedlemo:
The test for `RSVG::Handle` still fails. I will let you see where is the problem because you wrote the previous API. If you don't have the time, give me some hints in order to solve the issues.

I will create a PR later for the samples that are in *rsgv2-no-gi*.


rsvg2/lib/rsvg2/handle.rb
```diff
@@ -0,0 +1,49 @@
...
+    alias_method :initialize_raw, :initialize
+    def initialize(options={})
+      flags = options[:flags] || nil
+      file = options[:file] || nil
+      data = options[:data] || nil
+
+      if file
+        if flags
+          initialize_new_from_file(file)
+          flags = flags
```
kou: We need to use `initialize_new_from_gfile_sync()` to specify both `file` and `flags`.

rsvg2/lib/rsvg2/handle.rb
```diff
@@ -0,0 +1,49 @@
...
+        end
+      elsif data
+        initialize_new_from_data(data)
+      else
+        initialize_raw
+      end
+    end
+
+    # For backward compatibility
+    def new_from_data(data)
```
kou: It should be class method.

rsvg2/lib/rsvg2/handle.rb
```diff
@@ -0,0 +1,49 @@
...
+      new(:data => data)
+    end
+
+    # For backward compatibility
+    def new_from_file(file_name, options={})
```
kou: ditto.

rsvg2/test/test-handle.rb
```diff
@@ -69,17 +70,10 @@ def setup_large_svg
...
-      end
-
       def test_unlimited
         only_rsvg_version(2, 40, 3)
-        handle = RSVG::Handle.new_from_file(@large_svg_path,
-                                            :flags => :flag_unlimited)
+        handle = RSVG::Handle.new(:file => @large_svg_path,
+                                  :flags => :flag_unlimited)
         assert_equal([0, 0, 0.0, 0.0],
                      handle.dimensions.to_a)
```
kou: Umm. We need to rename `RSVG` module name to `Rsvg` module name to use the same namespace in `.gir`.

rsvg2/lib/rsvg2/handle.rb
cedlemo: > initialize_new_from_gfile_sync

Where does it come from ? I did not found anything related to this in the API documentation.

kou: It comes from https://developer.gnome.org/rsvg/stable/rsvg-Using-RSVG-with-GIO.html#rsvg-handle-new-from-gfile-sync

kou:
Thanks.

rsvg2/lib/rsvg2/handle.rb
cedlemo: Ha ok thanks. Sorry for the partial work.

kou: No problem!

PR #1040, opened 2017-06-19
Update glarea demo to new GLib::Bytes API

cedlemo:
I am not sure if it is the best way to handle it.


kou:
I've extend `GLib::Bytes` and fixed.

PR #1044, opened 2017-07-01
Ensure that there is no crash when the demo does not returns a widget

cedlemo:
Since `GtkPrintOperation` is a very peculiar widget  ( https://developer.gnome.org/gtk3/stable/gtk3-High-level-Printing-API.html) it seems that we can not returns a widget from the demo. So I just disable the font change in the demo selector (`TreeView` on the left of the UI) when the returned value does not have a `signal_connect` method.


kou:
Thanks.
I simplified your change. The demo just returns `nil` from `#run`.

PR #1047, opened 2017-07-01
Update to new api of GLib::Bytes

cedlemo:
I found this bug when I ran the markup.rb demo from Gtk3 gtk-demo : 

    ./main.rb -r markup


kou:
I've implemented it as needless `String` object creation.

Issue #1048, opened 2017-07-01
Pango missing constants.


cedlemo:
@kou, 
I can not find the following constants : 

* `Pango::AttrScale::XX_SMALL`
* `Pango::AttrScale::X_LARGE`

kou:
I've added them!

PR #1049, opened 2017-07-01
Update to new api of Pango

cedlemo:
It remains two constants to update but they are not implemented (see #1048 )


kou:
Thanks!!!

PR #1050, opened 2017-07-06
Update to new api of Pango

cedlemo:
@kou,

could you try this demo because it generates an error in `lib/gtk3/text-buffer.rb:182` 

It must be related to the GLib::Bytes new api : 

    NoMethodError: undefined method `bytesize' for #<GLib::Bytes:0x1ba6458 ptr=0x29fbd60 own=true>

Thanks.


kou:
Thanks. I've fixed.

PR #1065, opened 2017-07-20
check existence of Makefile before running make

dai-vdr:
this fix use other tests as reference.


kou:
When do we need this change?

dai-vdr:
This is needed when testing without building package.

Debian CI uses pre-built binary package (test target) and no-built source package (test case).
For example, see [ruby-cairo's test log](https://ci.debian.net/packages/r/ruby-cairo/unstable/amd64/). It does not run extconf.rb and does not generate Makefile.

kou:
OK. I've merged.

dai-vdr:
Thank you very much!

PR #1066, opened 2017-07-20
fix shebang path for examples

dai-vdr:
they are still using `/usr/local` as shebang path.


kou:
Thanks!
But we don't need to change `-no-gi` codes anymore.

dai-vdr:
Thank you.
Just FYI: I do not package `-no-gi`'s for Debian but `grep -r /usr/local` finds it.

PR #1071, opened 2017-07-21
add details about installing rsvg2 on heroku

Chocksy:


kou:
Thanks!
Could you use Markdown instead of RD?

We want to use Markdown for newly created ducuments.

Chocksy:
@kou i updated the file and included the latest updates too.

kou:
Thanks!

PR #1078, opened 2017-07-27
add generate gemspec task

dai-vdr:
I am working on Debian BTS: [ruby-gnome2: Please provide rubygem meta-information](https://bugs.debian.org/757464).
Antonio Terceiro <terceiro@debian.org> created this patch.
see:
https://lists.debian.org/debian-ruby/2017/07/msg00029.html
https://lists.debian.org/debian-ruby/2017/07/msg00030.html

### DIFF
--- a/glib2/lib/gnome2/rake/package-task.rb
+++ b/glib2/lib/gnome2/rake/package-task.rb
@@ -286,6 +286,13 @@ module GNOME2
       def define_package_tasks
         Gem::PackageTask.new(@spec) do |pkg|
         end
+        desc "Write #{@spec.name}.gemspec"
+        task :gemspec do
+          File.open("#{@spec.name}.gemspec", 'w') do |f|
+            f.write(@spec.to_ruby)
+          end
+          puts "#{@spec.name}.gemspec created"
+        end
       end
 
       class DependencyConfiguration


kou:
It seems that your patch already uses `dh_ruby`. Could you explain why you need `.gemspec`?

dai-vdr:
I thought I simply used `dh_ruby` to generate `.gemspec` from source at first, but it did not do.
So I read `dh_ruby` code and found it required _upstream created_ or _Debian maintainer prepared_ `.gemspec` to generate _package's_ `.gemspec`.
It means `dh_ruby` does not  _generate_ but _convert_ `.gemspec`.

https://anonscm.debian.org/cgit/pkg-ruby-extras/gem2deb.git/tree/lib/gem2deb/metadata.rb?h=0.35#n29
https://anonscm.debian.org/cgit/pkg-ruby-extras/gem2deb.git/tree/lib/gem2deb/metadata.rb?h=0.35#n104
https://anonscm.debian.org/cgit/pkg-ruby-extras/gem2deb.git/tree/lib/gem2deb/installer.rb?h=0.35#n74

kou:
Ah, are you using https://sourceforge.net/projects/ruby-gnome2/files/ruby-gnome2/ruby-gnome2-3.1.8/ruby-gnome2-all-3.1.8.tar.gz/download instead of `*.gem` to create `*.deb`?

We recommend that you use `*.gem` to create `*.deb`. `*.gem` includes `*.gemspec`.

Our `*.tar.gz` is provided only for old style system. We recommend that users including packagers use `*.gem` instead of `*.tar.gz`.

dai-vdr:
Thank you for your explain.  I understand it.
I will forward your message to Debian Ruby ML and discuss future ruby-gnome2 debs.

kou:
Thanks! If Debian packages have extra questions, please open a new issue.

terceiro:
FTR the reason Debian currently uses the source tarball is because ruby-gnome2 has been in Debian since before rubygems itself was a thing.

PR #1087, opened 2017-08-02
Vte3 regex

cedlemo:
@kou,

These Pull Request add support for `Vte::Regex`, the usage of `Glib::Regex` with `Vte::Terminal` is still working (with v = 0.48.3 at least) but is deprecated. (https://developer.gnome.org/vte/0.48/VteRegex.html).

I am sorry for the bad tests, I know that you consider that `assert_instance_of` are useless tests but I did not find anything else to test.


kou:
Thanks!

We need reader methods in `VteRegex` for tests. :<

PR #1099, opened 2017-09-03
Gtk3 sample gtk demo fishbowl rb

cedlemo:
It remains a little issue with the toggle button., the prevent changes toggle button should not appears at first.


kou:
Thanks!
I've fixed some issues.

PR #1103, opened 2017-09-14
Gtk3 sample gtk demo flowbox rb

cedlemo:
@kou ,

it seems that there is a problem with the memory allocation when you try to resize the window of the demo after launching it. Here is the error message:

```
./main.rb -r flowbox
/home/cedlemo/.gem/ruby/2.4.0/gems/gobject-introspection-3.1.9/lib/gobject-introspection/loader.rb:576: warning: GRClosure invoking callback: already destroyed: GtkDrawingArea::draw
/home/cedlemo/.gem/ruby/2.4.0/gems/gobject-introspection-3.1.9/lib/gobject-introspection/loader.rb:576: warning: GRClosure invoking callback: already destroyed: GtkDrawingArea::draw
/home/cedlemo/.gem/ruby/2.4.0/gems/gobject-introspection-3.1.9/lib/gobject-introspection/loader.rb:576: warning: GRClosure invoking callback: already destroyed: GtkDrawingArea::draw

...............

/home/cedlemo/.gem/ruby/2.4.0/gems/gobject-introspection-3.1.9/lib/gobject-introspection/loader.rb:576: [BUG] Segmentation fault at 0x00000000000038
ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-linux]

-- Control frame information -----------------------------------------------
c:0004 p:---- s:0025 e:000024 CFUNC  :invoke
c:0003 p:0107 s:0019 e:000018 LAMBDA /home/cedlemo/.gem/ruby/2.4.0/gems/gobject-introspection-3.1.9/lib/gobject-introspection/loader.rb:576 [FINISH]
c:0002 p:0371 s:0013 E:0021f0 EVAL   ./main.rb:518 [FINISH]
c:0001 p:0000 s:0003 E:0003f0 (none) [FINISH]

-- Ruby level backtrace information ----------------------------------------
./main.rb:518:in `<main>'
/home/cedlemo/.gem/ruby/2.4.0/gems/gobject-introspection-3.1.9/lib/gobject-introspection/loader.rb:576:in `block in define_method'
/home/cedlemo/.gem/ruby/2.4.0/gems/gobject-introspection-3.1.9/lib/gobject-introspection/loader.rb:576:in `invoke'

-- Machine register context ------------------------------------------------
 RIP: 0x00007fee3f84f8c4 RBP: 0x000000000233f5c0 RSP: 0x00007fffb94a12b0
 RAX: 0x0000000000755530 RBX: 0x0000000000000034 RCX: 0x0000000000000000
 RDX: 0x00007fffb94a1318 RDI: 0x000000000233f5c0 RSI: 0x000000000000009a
  R8: 0x0000000000000001  R9: 0x00000000006d2310 R10: 0x000000000233ff70
 R11: 0x00007fee3f7552f0 R12: 0x0000000000000c71 R13: 0x000000000233f5e8
 R14: 0x0000000000467e22 R15: 0x0000000000000001 EFL: 0x0000000000010246

-- C level backtrace information -------------------------------------------
```


kou:
Thanks!
We need to find lack of GC mark...

cedlemo:
@kou,

I did some tests with this : 

```ruby
diff --git a/gtk3/sample/gtk-demo/flowbox.rb b/gtk3/sample/gtk-demo/flowbox.rb
index 2e95368ac..7c85cd532 100644
--- a/gtk3/sample/gtk-demo/flowbox.rb
+++ b/gtk3/sample/gtk-demo/flowbox.rb
@@ -13,9 +13,9 @@ class ColorSwatch < Gtk::Button
   def initialize(color)
     super(:label => nil)
 
-    area = Gtk::DrawingArea.new
-    area.set_size_request(24, 24)
-    area.signal_connect "draw" do |_widget, cr|
+    @area = Gtk::DrawingArea.new
+    @area.set_size_request(24, 24)
+    @area.signal_connect "draw" do |_widget, cr|
       rgba = Gdk::RGBA.parse(color)
 
       if rgba
@@ -23,7 +23,7 @@ class ColorSwatch < Gtk::Button
         cr.paint
       end
     end
-    add(area)
+    add(@area)
     show_all
   end
 end
```

And everything seems to work. 

Does it means that the ruby value `area` is garbage collected when we don't set it as an instance variable and that when this ruby value is deleted, its closure (generated via `#signal_connect` is deleted too but not the underlying C widget (since they are always visible in the flowbox) ?

kou:
Yes.

Users should not care about it. We need more work to work the script without the changes.

I'll work on it after RubyKaigi 2017 (2017-09-18..2017-09-20) is finished. I have a talk in RubyKaigi 2017.

kou:
Sorry. I forgot this. :<
I've fixed this by marking all top level windows in GC.

cedlemo:
@kou, I am curious about how you find that the problem was related to the fact that the toplevel windows  where not marked for the GC ?

Did you use debugging tools ? did you make some tests ? what is something obvious for you?

kou:
First, I confirmed that what objects are GC-ed by adding the following code to the end of `flowbox.rb`:

```ruby
GLib::Timeout.add(1000) do
  p [
     :before,
     ObjectSpace.each_object(Gtk::Window) {},
     ObjectSpace.each_object(Gtk::DrawingArea) {},
     ObjectSpace.each_object(Gtk::ScrolledWindow) {},
     ObjectSpace.each_object(Gtk::FlowBox) {},
    ]
  GC.start
  p [
     :after,
     ObjectSpace.each_object(Gtk::Window) {},
     ObjectSpace.each_object(Gtk::DrawingArea) {},
     ObjectSpace.each_object(Gtk::ScrolledWindow) {},
     ObjectSpace.each_object(Gtk::FlowBox) {},
    ]
  GLib::Source::CONTINUE
end
```

All widgets were GC-ed.

Then, I thought that `main.rb` needs more code? For example, `main.rb` should keep all window references? But gtk3 gem users don't want to do it.

I remembered that we mark top-level windows in no-GI bindings: https://github.com/ruby-gnome2/ruby-gnome2/blob/master/gtk3-no-gi/ext/gtk3/rbgtk-window.c#L559-L564

This is a trick. :-)

cedlemo:
Thanks for taking the time to answers me. It maybe a trick for you but there are a lot of useful informations for me.

PR #1117, opened 2017-12-05
Use pkgconfig(foo) virtual provides for Red Hat based dependency

mtasaka:
As I wrote on https://github.com/ruby-gnome2/native-package-installer/issues/8 , Red Hat / Fedora package provides pkgconfig(foo) is such package contains foo.pc pkgconfig file. 

So instead of using package (rpm) name directory, using such virtual provides is preferred.

Note: I did not modify gnumeric/dependency-check/Rakefile this time, because this file does not contain ``unless PKGConfig.check_version?("foo")`` (yet?)

### DIFF
--- a/atk/dependency-check/Rakefile
+++ b/atk/dependency-check/Rakefile
@@ -32,7 +32,7 @@ namespace :dependency do
     unless PKGConfig.check_version?("atk")
       unless NativePackageInstaller.install(:altlinux => "libatk-devel",
                                             :debian => "libatk1.0-dev",
-                                            :redhat => "atk-devel",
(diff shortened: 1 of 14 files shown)


kou:
Thanks!

> Note: I did not modify gnumeric/dependency-check/Rakefile this time, because this file does not contain `unless PKGConfig.check_version?("foo")` (yet?)

Because `.pc` for Gnumeric doesn't exist on Debian. If it exists on Red Hat / Fedora, we can also use `pkgconfig(...)` approach for Gnumeric. Otherwise, we need to specify package name directly.

PR #1129, opened 2018-02-03
Gdk4 dev

cedlemo:
No great changes for now, I think that I will have to ask on the Gtk mailing list because as you can see in the links that I added in the README, they are doing a lot of changes (Remove all the stuff related to root windows etc, ...)


kou:
Thanks.
I've imported cairo related codes from gdk3.

kou:
> I will have to ask on the Gtk mailing list

It may be better that we migrate gtk-demo for GTK+ 4 to Ruby from C before we ask  on the GTK+ mailing list. We will be able to learn more about GTK+ 4 in the migration process.

cedlemo:
Oh yes, that is a good idea. My next PR will be the the gtk4 dir copied from gtk3 and a travis file like for Gdk4 so that I can work the gtk-demo from Gtk4 and adapt the gtk4 and the gdk4 as needed.

kou:
OK. But you don't need to include Travis related codes because Ubuntu 17.10 build already includes GTK+ 4 related packages.

Issue #1128, opened 2018-02-03
Gdk4 and Gdk::Atom


cedlemo:
@kou,

in Gdk4, it seems that the `Gdk::Atom` class is not automaticaly created, instead of that, all the related function for the atoms are functions of the Gdk4 module:

```ruby
puts Gdk.methods.grep(/atom/)
x11_atom_to_xatom
x11_atom_to_xatom_for_display                                                                                                                  
x11_get_xatom_by_name                                                                                                                          
x11_get_xatom_by_name_for_display                                                                                                              
x11_get_xatom_name     
x11_get_xatom_name_for_display                                                                                                                 
x11_xatom_to_atom 
x11_xatom_to_atom_for_display                                                                                                                 atom_intern                          
atom_intern_static_string   
atom_name 
``` 

Should we keep the same organization as before with a `Gdk::Atom`, if so should I try to map all the functions with a pattern (*atom*) with the loader.rb and create/fill the `Gdk::Atom` class there or should I
generate everything in the **gdk4/lib/gdk4/atom.rb** ?

kou:
It seems that `Gdk::Atom.intern` and `Gdk::Atom#name` work well.
We need more works just only for X11 relation atom functions.

I'll check it later.

cedlemo:
Well, I still have this with the jhbuilder environment that I just updated to be sure:

```bash
./gdk4_test.sh                                                                                                                                               Projets/Ruby
Traceback (most recent call last):
        15: from /home/cedlemo/Projets/Ruby/ruby-gnome2/gdk4/test/run-test.rb:56:in `<main>'
        14: from /home/cedlemo/.rvm/rubies/ruby-2.5.0/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
        13: from /home/cedlemo/.rvm/rubies/ruby-2.5.0/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
        12: from /home/cedlemo/Projets/Ruby/ruby-gnome2/gdk4/lib/gdk4.rb:33:in `<top (required)>'
        11: from /home/cedlemo/Projets/Ruby/ruby-gnome2/gdk4/lib/gdk4.rb:41:in `<module:Gdk>'
        10: from /home/cedlemo/Projets/Ruby/ruby-gnome2/gdk4/lib/gdk4/loader.rb:21:in `load'
[...]

PR #1134, opened 2018-02-15
Gtk4 first commit

cedlemo:
I removed most of the Gtk3 related stuff (ext/ ) exept libs in *gtk4/lib/gtk4/* because I will need them latter.

The test for version is working.

@kou,
~~is it possible that we disable the loading of the atom.rb in Gdk4 (and not tests it)? Both the Gtk4 libs from jhbuilder or from my system (ArchLinux) don't implement anymore `GdkAtom`. In order to test both Gtk4 or Gdk4, I have to either comment the line `require "gdk4/atom"` or commit and use a docker instance locally or via Travis which take a lot of resources and or  time.~~

Oups I forgot we already discuss this in #1128

### DIFF
--- a/gtk4/COPYING.LIB
+++ b/gtk4/COPYING.LIB
@@ -1,0 +1,502 @@
+                  GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 2.1, February 1999
+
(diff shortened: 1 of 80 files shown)


kou:
Thanks!

PR #1135, opened 2018-02-15
Gtk4 sample hello world

cedlemo:
First example. I created this branch from the gtk4 branch that I pushed previously.


kou:
Thanks!!

PR #1144, opened 2018-02-28
Gtk4 sample utils

cedlemo:
I added a utils.rb file in order to share a function to load the Gtk4/Gdk4 libs while there are no gems for them.


kou:
Thanks!

PR #1161, opened 2018-04-24
change pixels-above-lines to pixels_above_lines in Decorative Overlay…

kojix2:
Hi!
I think the original code may be correct.
However, it did not work on my computer.
Good night.


kou:
Thanks.
I've improved `Gtk::TextBuffer`. `Gtk::TextBuffer` should also accept `pixels-above-lines` style.

PR #1173, opened 2018-05-07
Gtk4 : Customize Gtk::Box#new

cedlemo:
I just kept the rewrite of `Gtk::Box.new` from the Gtk3 file, the API changed for : 

* https://developer.gnome.org/gtk4/3.93/GtkBox.html#gtk-box-set-child-packing
* https://developer.gnome.org/gtk4/3.93/GtkBox.html#gtk-box-pack-start


kou:
Thanks!

PR #1174, opened 2018-05-08
Update project website domain

okkez:
http://ruby-gnome2.sourceforge.jp ->
https://ruby-gnome2.osdn.jp

Redirect to http://ruby-gnome2.osdn.jp from http://ruby-gnome2.sourceforge.jp

I don't update URL in the NEWS file because it is the history.


kou:
Thanks!

PR #1186, opened 2018-05-19
gtk3 sample: work popover demo with GTK+ 3.18 again

kojix2:
Thank you for your update. 
I checked https://github.com/ruby-gnome2/ruby-gnome2/issues/1184 and found it still did not work.
I add popover-3.18.ui to demo.gresource.xml.
Please check it.


kou:
Good catch!

Issue #1182, opened 2018-05-18
Error when loading gschema with wayland backend.


cedlemo:
I have an application that load settings from a custom gschema like described in https://github.com/ruby-gnome2/ruby-gnome2/tree/master/gtk3/sample/tutorial#define-and-store-settings-for-an-application-with-gschemas and I have a strange issue with the wayland backend:

I know it is a wayland backend because of the following environment variables: `$DISPLAY=:0` and `$WAYLAND_DISPLAY=wayland-0`.

 When I run my application that has been installed as a gem:

```
topinambour                                                               ~

(topinambour:7303): GLib-GIO-ERROR **: 20:15:07.218: Settings schema 'com.github.cedlemo.topinambour' is not installed

[1]    7303 trace trap (core dumped)  topinambour
```

I have the same error when I launch the application in its git local copy with `./bin/topinambour`

In both I tried to use : 
```
Gio::SettingsSchemaSource.new(data_path,
                              Gio::SettingsSchemaSource.default,
                              false)
``` 
instead of using `ENV[GSETTINGS_GSCHEMA_DIR]= data_path` but it did not works neither.

When I run in a local copy of ruby-gnome2 the sample *gtk3/sample/tutorial/exampleapp10* with `ruby exampleapp.rb` it works.

If I set in my .zshrc `export GSETTINGS_GSCHEMA_DIR=a/path/` it works.

I checked with a classical Xorg session : `$DISPLAY=:0` and `$WAYLAND_DISPLAY=` and I can run my application both from the gem installation or from the local copy of the git repository.

Any ideas ?

kou:
How do you instantiate your `Gio::Settings`?
You need to specify the settings schema source:

```ruby
source = Gio::SettingsSchemaSource.new(data_path,
                                       Gio::SettingsSchemaSource.default,
                                       false)
settings = Gio::Settings.new(source)
```

cedlemo:
`Gio::Settings#new` only accepts a string as argument: 

```
/home/cedlemo/.gem/ruby/2.5.0/gems/gobject-introspection-3.2.6/lib/gobject-introspection/loader.rb:317:in `invoke': no implicit conversion of Gio::SettingsSchemaSource into String
	 from /home/cedlemo/.gem/ruby/2.5.0/gems/gobject-introspection-3.2.6/lib/gobject-introspection/loader.rb:317:in `block (2 levels) in load_constructor_infos'
[...]

PR #1187, opened 2018-05-19
Gio2 settings new

cedlemo:
#1182


kou:
Thanks!
I changed to accept `Gio::Settings.new(:schema => schema)`.

PR #1212, opened 2018-05-31
Clutter-gdk : add static gem

cedlemo:
@kou, I removed the `dependency-check` references in the *gemspec* file. I am not sure about this.


kou:
Thanks!!
It doesn't need `dependency-check` because required `.gir` is provided by Clutter. Check for Clutter is done by depended clutter gem.

PR #1235, opened 2018-08-03
test-source: Don't assert anything about the idle source's initial name

iainlane:
This is not defined. It was previously empty, but as it happens GLib
2.57.2 started setting default names for its sources. We should instead
test that we can set a name on a source.

See: https://gitlab.gnome.org/GNOME/glib/commit/208a6e815affead7ef8ed86439fcc2bb6a86d56a


kou:
Good catch!

PR #1253, opened 2018-11-07
use RSVG instead of Rsvg

kojix2:
fix typo
https://github.com/ruby-gnome2/ruby-gnome2/issues/1252

### DIFF
--- a/rsvg2/rsvg2.gemspec
+++ b/rsvg2/rsvg2.gemspec
@@ -20,8 +20,8 @@ require_relative "../glib2/version"
 
 Gem::Specification.new do |s|
   s.name          = "rsvg2"
-  s.summary = "Ruby/Rsvg is a Ruby binding of librsvg-2.x."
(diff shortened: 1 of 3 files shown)


rsvg2/rsvg2.gemspec
```diff
@@ -20,8 +20,8 @@ require_relative "../glib2/version"
 
 Gem::Specification.new do |s|
   s.name          = "rsvg2"
-  s.summary = "Ruby/Rsvg is a Ruby binding of librsvg-2.x."
-  s.description = "Ruby/Rsvg is a Ruby binding of librsvg-2.x."
+  s.summary = "Ruby/RSVG is a Ruby binding of librsvg-2.x."
+  s.description = "Ruby/RSVG is a Ruby binding of librsvg-2.x."
```
kou: Can you append "2"?

kojix2: OK. I will append 2.

kou:
Thanks!

PR #1259, opened 2018-11-21
update gtk3 sample/misc calendar

kojix2:
1. remove "Gtk::CalendarDisplayOptions::NO_MONTH_CHANGE" so that you can change month.
2. use `year`, `month`, `day` properties in GtkCalendar.


kou:
Thanks!

PR #1260, opened 2018-11-21
update gtk misc sample flowbox

kojix2:
1. use Cairo context instead of `Gtk::Widget#override_background_color` that is deprecated since v3.16. 
 https://github.com/ruby-gnome2/ruby-gnome2/issues/1080
2. Other changes

### DIFF
--- a/gtk3/sample/misc/flowbox.rb
+++ b/gtk3/sample/misc/flowbox.rb
@@ -1,7 +1,7 @@
 #!/usr/bin/env ruby
 #
 # Copyright (c)  2014  Gian Mario Tagliaretti
-# Copyright (c)  2015  Ruby-GNOME2 Project Team
+# Copyright (c)  2015-2018  Ruby-GNOME2 Project Team
 #
 # Permission is granted to copy, distribute and/or modify this document
 # under the terms of the GNU Free Documentation License, Version 1.3
@@ -37,7 +37,7 @@ class FlowBoxWindow < Gtk::Window
     set_default_size(300, 250)
 
     header = Gtk::HeaderBar.new
-    header.set_title("Flow Box")
+    header.title = "Flow Box"
     header.subtitle = "Sample FlowBox app"
     header.show_close_button = true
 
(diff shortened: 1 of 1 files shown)


gtk3/sample/misc/flowbox.rb
```diff
@@ -46,37 +46,36 @@ def initialize
...
-  def color_swatch_new(color_name)
-    color = Gdk::RGBA.parse(color_name)
+  def create_color_swatch(color_name)
     button = Gtk::Button.new
 
     area = Gtk::DrawingArea.new
     area.set_size_request(24, 24)
-    area.override_background_color(0, color)
+    area.signal_connect("draw") do |_, context|
+      color = Gdk::RGBA.parse(color_name)
```
kou: It's better that we parse color name outside of the signal handler. Because signal handler is called multiple times but `create_color_swatch` is called only once.

gtk3/sample/misc/flowbox.rb
```diff
@@ -46,37 +46,36 @@ def initialize
     scrolled = Gtk::ScrolledWindow.new
     scrolled.set_policy(Gtk::PolicyType::NEVER, Gtk::PolicyType::AUTOMATIC)
 
-    @flowbox = Gtk::FlowBox.new
-    @flowbox.set_valign(Gtk::Align::START)
-    @flowbox.set_max_children_per_line(30)
-    @flowbox.set_selection_mode(Gtk::SelectionMode::NONE)
-    fill_flowbox
+    flowbox = Gtk::FlowBox.new
+    flowbox.set_valign(Gtk::Align::START)
```
kou: Can we use `flowbox.valigh = :start`?

gtk3/sample/misc/flowbox.rb
```diff
@@ -46,37 +46,36 @@ def initialize
...
+    flowbox.set_max_children_per_line(30)
```
kou: Can we use `flowbox.max_children_per_line = 30`?

gtk3/sample/misc/flowbox.rb
```diff
@@ -46,37 +46,36 @@ def initialize
...
+    flowbox.set_selection_mode(Gtk::SelectionMode::NONE)
```
kou: Can we use `flowbox.selection_mode = :none`?

gtk3/sample/misc/flowbox.rb
kojix2: all right. Move this line up.

kou:
Thanks!

PR #1268, opened 2019-01-05
fix typos

kojix2:


kou:
Good catch!
We don't need to maintain `*-no-gi` directories. They just exist only for easy to refer (for me).

PR #1280, opened 2019-04-30
Fix title link

ismailarilik:


kou:
Good catch!

ismailarilik:
There are other similar issues, too.

PR #1281, opened 2019-05-16
Fix indentation of Table of Contents

ismailarilik:


kou:
Thanks!
Could you change to use 1 space instead of 2 spaces after `*` such as `*  XXX` -> `* XXX`?

kou:
Thanks!

Issue #1294, opened 2019-09-05
glib-enum-types.c generated with GLib 2.61.3 fails to compile


mtasaka:
As written on summary, glib2/ext/glib2/glib-enum-types.c generated with GLib 2.61.3 fails to compile like:
```
DEBUG: glib-enum-types.c: In function 'g_test_subprocess_flags_get_type':
DEBUG: glib-enum-types.c:729:12: warning: initialization of 'unsigned int' from 'const char *' makes integer from pointer without a cast [-Win
t-conversion]
DEBUG:   729 |       { }, "}", "}" },
DEBUG:       |            ^~~
DEBUG: glib-enum-types.c:729:12: note: (near initialization for 'values[4].value')
DEBUG: glib-enum-types.c:729:12: error: initializer element is not computable at load time
DEBUG: glib-enum-types.c:729:12: note: (near initialization for 'values[4].value')
DEBUG: glib-enum-types.c:729:7: warning: missing initializer for field 'value_nick' of 'GFlagsValue' {aka 'const struct _GFlagsValue'} [-Wmiss
ing-field-initializers]
DEBUG:   729 |       { }, "}", "}" },
DEBUG:       |       ^
```

mtasaka:
The generated file (with GLib 2.61.3) is uploaded at https://mtasaka.fedorapeople.org/Bugfix/ruby-gnome/issue1294/glib-enum-types.c .
There is some change on glib/gtestutils.h in https://gitlab.gnome.org/GNOME/glib/commit/e440249bd3c94ec9553c5617b152abf0bb575767 and it seems that glib-mkenums.rb fails to parse new gtestutils.h header (I guess).

mtasaka:
Actually the related change in GLib is this one: 
https://gitlab.gnome.org/GNOME/glib/commit/40ff4759774516b7dff3b07303a98fc1626b2d0e

PR #1295, opened 2019-09-06
glib2: fix parser in MkEnums for resent GLib

mtasaka:
GLib 2.62.0 introduced some deprecation warnings for macros in gtestutils.h with the commit https://gitlab.gnome.org/GNOME/glib/commit/40ff4759774516b7dff3b07303a98fc1626b2d0e .
This patch makes GLib::MkEnums correctly parse header files in GLib 2.62.0 again.

Fixes https://github.com/ruby-gnome/ruby-gnome/issues/1294


kou:
Thanks for your pull request.
I'm preparing test environment. Please wait for while.

kou:
Thanks!

PR #1297, opened 2019-09-16
Move non-gobject-introspection directories to no-gi

kojix2:
Hello.
* Move no-gi directories to no-gi directory.

This pull request is not to improve the code, but rather to suggest a website design.
Improve human eye/brain search performance by optimizing strings.

image

image

This is just one suggestion. Please close it if there is a practical problem.


kou:
If we move them to non top-level directory, we can't build them anymore. They depend on `../glib2/`.

kojix2:
My bad.
I don't really understand how to build ruby-gnome.
Although this library is maintained by your special abilities and mechanisms, I still think too many directory files at the top level is a problem. 
I will try to understand how ruby gnome is built.

Thank you!

PR #1302, opened 2019-10-10
Fix typo

kojix2:
Maybe typo?

> NameError: undefined local variable or method `atk' for main:Object


kou:
Thanks. It's a typo but we should use `package_id`.

PR #1309, opened 2019-12-08
Add Ubuntu package name for gtksourceview4

ismailarilik:
It is `libgtksourceview-4-dev`, not `libgtksourceview-4.0-dev` as for Debian.


kou:
Thanks!

PR #1393, opened 2020-04-29
rsvg: fix usage of deprecated GDK Window size to width, height

JC-LL:
Fixes usage of deprecated method Gdk window "size" in svg-viewer example.


rsvg2/sample/svg-viewer.rb
```diff
@@ -20,7 +20,7 @@
 area = Gtk::DrawingArea.new
 area.signal_connect(:draw) do |widget, _event|
   context = widget.window.create_cairo_context
-  window_width, window_height = widget.window.size
+  window_width, window_height = widget.window.width,widget.window.height
```
kou, line 23:
```suggestion
  window_width, window_height = widget.window.width, widget.window.height
```

PR #1416, opened 2020-12-18
Removed unused travicCI badge

kojix2:
* Remove unused TravisCI badges
* Move the badge position to below the title

I considered adding a badge for Github Actions.
There are three separate badges for Linux, Mac, and Windows.
If it were my personal project, I would add the badges. 

But in the case of ruby-gnome, I can't decide if I should add badges. 
If there are too many badges, it doesn't look elegant.

I don't care about the badge itself, 
but seeing a badge that has stopped working is not a good feeling.

Thank you very much.


kou:
Thanks!

I don't have any opinion about badges because I don't see badges,

PR #1428, opened 2021-06-06
Fix pango metadata

brodock:
While trying to figure out where pango's sourcecode was, by looking at https://rubygems.org/gems/pango/ I just couldn't. This adds extra links to the rubygems metadata that will populate rubygems with those.

I've also defined minimum supported ruby version (as per current changelog it seems we support `>=2.6` only).

You can learn more about the rubygem specification bits I've modified here: https://guides.rubygems.org/specification-reference/

### DIFF
--- a/pango/pango.gemspec
+++ b/pango/pango.gemspec
@@ -21,7 +21,7 @@ require_relative "../glib2/version"
 Gem::Specification.new do |s|
   s.name          = "pango"
   s.summary       = "Ruby/Pango is a Ruby binding of pango-1.x."
-  s.description   = "Ruby/Pango is a Ruby binding of pango-1.x."
+  s.description   = "Ruby/Pango is a Ruby binding of pango-1.x based on GObject-Introspection."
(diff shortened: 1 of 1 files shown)


pango/pango.gemspec
```diff
@@ -39,10 +39,17 @@ Gem::Specification.new do |s|
   ]
   s.files += Dir.glob("lib/**/*.rb")
   s.files += Dir.glob("ext/#{s.name}/*.{c,h,def,rb}")
-  s.files += Dir.glob("test/**/*")
 
   s.add_runtime_dependency("cairo-gobject", "= #{s.version}")
   s.add_runtime_dependency("gobject-introspection", "= #{s.version}")
+  s.required_ruby_version = '>= 2.6'
```
kou, line 45: Could you revert this?
We can't maintain this. We specify this only when we dropped support for not EOL-ed Ruby. (We'll not do it.)

brodock: done

kou:
Thanks.

PR #1435, opened 2021-09-14
Fix pkgconfig id for gdk4

hellcp:
pkgconfig id for gdk4 is the same as gtk4
https://koji.fedoraproject.org/koji/rpminfo?fileStart=350&rpmID=27692196#filelist


pterjan:
Thank you, this is indeed the correct name

Issue #1455, opened 2022-01-13
[Error / Possible bug] gobject-introspection/loader.rb:555:in `define_method': can't modify frozen Hash: {"create_pango_context"= [and a lot more output here]


rubyFeedback:
Hey there kou,

I just did "gem install gtk4" and then "gem install gtk3" to make sure I have the latest two gems installed.

I then downloaded via git clone the latest source code of ruby-gnome here.

Now, I navigate to this directory:

    cd /Depot/jjjj/ruby-gnome/gtk4/sample/

In that directory I run:

    ruby drawing.rb

I then get exactly this error:

    /Depot/jjjj/ruby-gnome/gobject-introspection/lib/gobject-introspection/loader.rb:555:in `define_method':
    can't modify frozen Hash: {"create_pango_context"= [and a lot more output here]

Not entirely certain where the error is but I suppose there are not that many instances of hash usage in that .rb file. In fact,
in the .rb file I can not seem to find any Hash at all, so perhaps gobject-introspection does something "behind the scene".

(Prior to that I had a ractor-related error, but that error went away after I did the two "gem install" commands.)

rexkogitans:
@rubyFeedback I have exactly the same problem. Did you solve it?

HuBandiT:
Temporary workaround: #1474

kou:
Fixed.

PR #1474, opened 2022-05-06
GTK4: work around frozen and/or missing INVOKERS

HuBandiT:
Workaround to allow GTK4 to load despite some INVOKERS being frozen (Cairo::Context) or outright missing (Cairo::Surface).

Fixes the symptoms reported in #1455 but does not look at the actual cause, so it is not recommended for merging without further verification/investigation.


kou:
Sorry for not working on this.
I'll update gdk4 and gtk4 with proper way.

HuBandiT:
Your work honours us. Thank you in advance.

If there is documentation on the architecture, maybe some of us can help with small bits of work, now or in the future.

HuBandiT:
... or if you can describe what needs to be done in this case.

kou:
You're right. I should prepare documentation... But I don't like/good at writing documentation (in English)...

Idea: I can explain the architecture in Japanese verbally without pain. If someone listens the explanation in Japanese and writes documentation in English, I'm happy. :-) 

For this case, we need to re-prepare gdk4/gtk4 based on the latest gdk3/gtk3 because the current gdk4/gtk4 are based on old gdk3/gtk3 and we have many improvements in gdk4/gtk4 after the gdk4/gtk4 creations. I'm doing it for gdk4 for now.

HuBandiT:
Sadly, I do not understand spoken Japanese. But I am reasonable in writing English technical text. So if you can explain in written Japanese (which I can try to understand using translation services), or in very basic English writing... and in code... then I (and others with time and skills to edit documentation but no knowledge of the architecture) can write better English documentation. A sharing of tasks/work: you give the information with minimal effort, others give good language (text editing).

kou:
I found a person who can listen Japanese and write explained description to documentation in Japanese. We will try creating documentation a bit in this month. If we can create documentation, please translate it to English and brush up the translated documentation.

HuBandiT:
That is exciting news! Yes, let's try! It will be difficult in the beginning. But I hope we can get something done. And if we share it publicly, I hope others will join in as well.

kou:
Please try the current master.

HuBandiT:
Thank you, it works now.

PR #1478, opened 2022-06-04
Fix audio-example

kojix2:
* Use [pipeline_clock](https://github.com/GStreamer/gstreamer/blob/2db3ddaa9d8e9e0a1659d7e48623c512dbda691b/subprojects/gstreamer/tests/examples/controller/audio-example.c#L30)

```ruby
bin.clock              # => nil
bin.pipelie_clock      # => Gst::SystemClock
```

* Use warn instead of $stderr.puts
  * If you do not like this change, undo and merge.

### DIFF
--- a/gstreamer/sample/audio-example.rb
+++ b/gstreamer/sample/audio-example.rb
@@ -32,7 +32,7 @@
 require "gst"
 
 bin = Gst::Pipeline.new("pipeline")
-clock = bin.clock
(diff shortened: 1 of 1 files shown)


kou:
Thanks.

I prefer `$stderr.puts` to `warn` in this case because `$stderr.puts` always outputs even if `$VERBOSE` is `nil`.

PR #1482, opened 2022-06-18
Documentation tasks for Rakefile

spchamp:
The changeset in this pull request will add documentation tasks to the Rakefile
- Task `gem:docs` as presented with `rake -T`, using both of the following tasks
- Task `gem:girdocs` for documentation using yard-gobject-introspection
- Task `gem:rubydocs` for other Ruby docs, using yard

Documentation files will be produced in subdirectories under ./yard_docs/
- in `yard_docs/gir/` for docs using yard-gobject-introspection
- in `yard_docs/ruby/` for Ruby docs using yard

`./yard_docs/` will be removed under the `clean` Rakefile task

`.gitignore` is updated for files created by yard

In commentary: 

Perhaps these Rakefile tasks may serve as any utility for local access to the Ruby-GNOME documentation. This would not be integrated with the documentation tasks for GitHub Actions in Ruby-GNOME.

I'm not certain if all of the decisions in this changeset would be ideal for the project. 

For these added tasks, I've used the `yard_docs` path under the project root directory. This path has been used in the GitHub Actions for Ruby-GNOME. 

The Rakefile tasks added in this changeset would output the documentation to individual directories under  `./yard_docs/`, respectively `gir` for the documentation generated with the yard-gobject-introspection extension for yard, and `ruby` for the rest of the documentation generated by yard.

The `gem:rubydocs` task will add `*.h` files to the set of files processed for the documentation. Although at this time, I'm not personally well familiar with any conventions for developing Ruby extensions, it seems yard was able to process these C header files for the documentation.

While it may be a trivial contribution, I wanted to share this patch along to upstream here. Perhaps the added Rakefile tasks may serve as any kind of helpful utility, whether as presented in this changeset or with any adaptation.

**Update:** The second changeset moves the 'clean' task handling for yard_docs into the main  `clean` task, not requiring `rake/clean` and its side effects for the `clean` task (e,g would remove editor backup files in clean). Simply using `rm -rf` for the yard_docs dir in `clean`, no additional files will be removed with this changeset.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.o
 *.so
 *.swp
+*.yardoc
 .emacs.desktop
 .emacs.desktop.lock
 Makefile
@@ -27,6 +28,7 @@ mkmf.log
 */tmp/
 */vendor/
 */pkg/
+*/yard_docs/
 .gdb_history
 /dist/
 /misc/release.rb
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@
 *.o
 *.so
 *.swp
-*.yardoc
 .emacs.desktop
 .emacs.desktop.lock
 Makefile
@@ -37,5 +36,6 @@ yard_docs
 /Gemfile.lock
 /Gemfile.local
 /build/.vagrant/
+/.yardoc/
 .RUBYLIBDIR*.time
 .RUBYARCHDIR.time
--- a/.gitignore
+++ b/.gitignore
@@ -28,7 +28,7 @@ mkmf.log
 */tmp/
 */vendor/
 */pkg/
-*/yard_docs/
+yard_docs
 .gdb_history
 /dist/
 /misc/release.rb
--- a/Rakefile
+++ b/Rakefile
@@ -93,7 +93,7 @@ end
 desc "clean all packages, docs"
 task :clean do
   sh("make", "clean") if File.exist?("Makefile")
-  sh("rm", "-rf", "./yard_docs")
+  rm_rf("yard_docs")
 end
 
 desc "more clean all packages"
@@ -128,7 +128,7 @@ namespace :gem do
                                 ygir_gem.full_gem_path)
     args = [ygir_lib, * packages_rb]
     all_args = %w(bundle exec yard doc -o ./yard_docs/gir/ --load).concat(args)
-    sh Shellwords.join(all_args)
+    sh(*all_args)
   end
 
   task :rubydocs do
@@ -138,7 +138,7 @@ namespace :gem do
         File.join(pkg, "ext", pkg, "**/*.h")
       ]}
     all_args = %w(bundle exec yard doc -o ./yard_docs/ruby/).concat(file_globs)
-    sh Shellwords.join(all_args)
+    sh(*all_args)
   end
 
   desc "push all gems"
--- a/Rakefile
+++ b/Rakefile
@@ -4,6 +4,7 @@ require "pathname"
 require "find"
 require "tmpdir"
 require "open-uri"
+require 'shellwords'
 
 task :default => :test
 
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -3,6 +3,7 @@
 *.o
 *.so
 *.swp
+*.yardoc
```
kou, line 6: What paths are you expected by this pattern?

I think that `/.yardoc/` is correct entry.

.gitignore
```diff
@@ -27,6 +28,7 @@ mkmf.log
 */tmp/
 */vendor/
 */pkg/
+yard_docs
```
kou, line 31: How about using `/docs/` instead of `yard_docs`?
We generate our docs only by YARD.

Rakefile
```diff
@@ -89,9 +90,10 @@ task :build => ["Makefile"] do
   sh("make")
 end
 
-desc "clean all packages"
+desc "clean all packages, docs"
 task :clean do
   sh("make", "clean") if File.exist?("Makefile")
+  sh("rm", "-rf", "./yard_docs")
```
kou, line 96:
```suggestion
  rm_rf("yard_docs")
```

Rakefile
```diff
@@ -114,6 +116,31 @@ namespace :gem do
...
+  task :girdocs do
+    packages_rb = packages.map { |pkg|
+      File.join(pkg, "lib", pkg + ".rb")
+    }
+    ygir_gem = Gem::Specification.find_by_name('yard-gobject-introspection')
+    ygir_lib = File.expand_path('lib/yard-gobject-introspection.rb',
+                                ygir_gem.full_gem_path)
+    args = [ygir_lib, * packages_rb]
+    all_args = %w(bundle exec yard doc -o ./yard_docs/gir/ --load).concat(args)
+    sh Shellwords.join(all_args)
```
kou, line 131: We don't need to use `Shellwords` because `sh` accepts variable size arguments:

```suggestion
    sh(*all_args)
```

Rakefile
```diff
@@ -114,6 +116,31 @@ namespace :gem do
     end
   end
 
+  desc "document all gems"
+  task docs: [:girdocs, :rubydocs]
```
kou, line 120: I'm OK with separated docs as the first step.
But I want to mix `yard-gobject-introspection` based documents and `*.rb`/`.h`/`*.c` based documents eventually.

spchamp: It might be possible for yard-gobject-introspection to be used a plugin for yard?

I was looking recently at the structure of the Invoker, FunctionInfo and virtual function structures in loader.rb for gobject-introspection. Perhaps these could be used for merging the gobject introspection metadata and its implementation in Ruby into the YRI metadata structures used in Yard. 

I believe that Yard's YRI data may be similar to the RI structures in rdoc. Ostensibly, this YRI data could be used directly for documentation browser for the desktop, as well as for any role in producing a merged HTML documentation. I'll try to take a look at the Yard API here.

First, to update the patch for the Rakefile though

.gitignore
spchamp: There are [files in a `doc` directory](https://github.com/ruby-gnome/ruby-gnome/tree/master/doc), for Heroku installation with rsvg2 and Release instructions. For the `yard_docs` pathname, it would not be ambiguous with the `doc` directory. 

In [`.github/workflows/doc.yml`](https://github.com/ruby-gnome/ruby-gnome/blob/master/.github/workflows/doc.yml), a  `yard_docs` dir is used in the `bundle exec yard doc` step, before renaming the directory to a location for publishing the documentation. This directory would have a different structure than the `yard_docs` dir that would be generated with the Rakefile in the patch. It seemed like a common pathname that was distinct to the pathname for the project documentation

kou: The `yard_doc` name wasn't considered carefully.

Anyway, I'll merge this as-is. We can organize directory struct as follow-up tasks.

spchamp: fwiw for the idea of combining the documentation for Gir definitions and their implementation in RG with the main Yard docs for RG: I think Yard process the source code, when generating docs? 

I may have been able to figure out some aspects for type mapping when ruby-gnome is already loaded, e.g for classes that provide an `INVOKERS` constant and/or `INITIALIZE_INVOKERS` and for modules that provide an `INVOKERS` constant under the module's singleton class. If  a Yard plugin would be able to load the Ruby-GNOME API for which it's generating documentation, it could process those values to generate documentation for the method mapped to each Invoker instance?

To iterate on these values, the documentation plugin could need to be provided with a module name, where it would begin iteration, ideally after Yard has processed the source definitions.

This itself would not provide complete API coverage? For instance, definitions provided e.g any dynamically loaded definitions under `GLib` might not be reached if the plugin would only be processing for Invoker instances?

If the plugin was provided with a module name or a list of module names, or if the module names could be computed from what Yard has processed, perhaps there may be a way to run the API walker after Yard has already processed the source code. For each class or module definition beginning at each of the named modules, if  Yard has not found documentation for that class, module, and/or any public (or other) method on the same, the API walker may be able to add documentation from any definitions located via Ruby GNOME?

If I can figure out how the plugin might iterate on any dynamically loaded definitions (??) under GLib, hopefully then it could provide complete API coverage, with tests. Then, perhaps there would only be a question of how to produce something that Yard could use for its YRI structures then adding that to Yard's parse, in a transformation on each definition?

kou: How about creating a new issue  https://github.com/ruby-gnome/ruby-gnome/issues/new or new discussion https://github.com/ruby-gnome/ruby-gnome/discussions/new for further discussion instead of reusing a closed pull request?

PR #1484, opened 2022-06-20
Args patch for Vte::Terminal#feed_child

spchamp:
With Ruby-GNOME using introspection data from GLib 2.70.4, then using the original `Vte::Terminal feed_child` method aliased as `feed_child_raw`, the `feed_child_raw` method was not accepting two args.

When `Vte::Terminal#feed_child_raw` was called with one arg, a string, then it would send the string to the Pty

### DIFF
--- a/vte3/lib/vte3/terminal.rb
+++ b/vte3/lib/vte3/terminal.rb
@@ -64,10 +64,5 @@ module Vte
         end
       end
     end
-
-    alias_method :feed_child_raw, :feed_child
-    def feed_child(text)
-      feed_child_raw(text, text.bytesize)
-    end
   end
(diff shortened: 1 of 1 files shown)


kou:
It's needed for old VTE. It seems that GObject Introspection annotation was fixed by https://gitlab.gnome.org/GNOME/vte/-/commit/00f252725c1eb1c5dc65fd9901221523418e18be . The fix is included in 0.45.90 or later.

Ubuntu 18.04 ships VTE 0.52.1: https://packages.ubuntu.com/search?keywords=libvte

OK. We drop support for VTE 0.44.90 or earlier.

PR #1490, opened 2022-07-26
gio2, gtk3: defining Gio Action support with new rbgobj_register_mark_func handlers

spchamp:
For a purpose of a sense of _widget-scoped actions_ in a UI design for an application, I've been using a number of custom **Gio::SimpleActionGroup** to provide custom action prefixes other than `win` or `app`. Each custom action group is bound to some individual widget in the UI. 

This has been approached as so, as in order to call an action's callback function from some **Gtk::Actionable** widget within the effective _widget scope_ of each bound **Gio::SimpleAction**

The implementation of each action in Ruby would correspond, ideally, to any single _Action Name_ entered in the Glade UI design for a **Gtk::Actionable** widget e.g `prefix.name` or `prefix.name::detailed-part`. After coordinating the Glade UI design with the implementation in Ruby, the custom action prefix/name bindings have been working out up to the point of GC. 

After GC, I'd been seeing the following message for actions defined under these custom action prefixes or action groups:

~~~~
Ruby/GLib2-WARNING **: 17:04:48.170: GRClosure invoking callback: already destroyed: GSimpleAction::activate
~~~~

The GTK UI would then present each actionable widget as though each such action was still available, but the widget would not activate the action's `activate` signal callback function after GC. This was not encountered for an action bound under the `app` prefix, e.g the `app.quit` action was still available. It was encountered for actions bound under other prefixes, e.g `editor.popup`

In trying to develop a patch for this, I've found an approach that seems to have mostly worked out. With the patch, that message no longer appears after GC. Moreover, each action group then continues to be available for facilitating the activation of any individual action via the GTK UI.

I'm not certain if more of an illustration could be helpful for explaining the rationale of this methodology. The following patch seems to work. Quoting the changelog description:

rb-gio2.c:
- defining mark functions for Gio::Action objects encapsulated in each Gio::SimpleActionGroup or Gio::Application
- Declaring rbgobj_register_mark_func handlers for Gio::SimpleActionGroup, Gio::Application, and Gio::Action in Init_gio2

rb-gtk3-widget.c:
- Defining a mark function for marking each encapsulated GActionGroup in each Gtk::Widget
- Declaring rb_gtk3_widget_mark as an rbgobj_register_mark_func handler in rbgtk3_widget_init

rb-gtk3.h:
- Adding a reference to the gio header

gtk3.gemspec:
- Adding gio as a dependency

test-gtk-widget.rb
- Adding a test case for Gio::ActionGroup encapsulation in Gtk::Widget

----

**Known limitations to this patch:** 

Given the following in the description for `gtk_widget_get_action_group` and concerning the mark function added to **Gtk::Widget** with this patch, I'm afraid that the Gtk::Widget mark function may mark an action group more than once 

>  "The resulting GActionGroup may have been registered to widget or any GtkWidget in its ancestry." 
[[Gtk.Widget.get_action_group](https://docs.gtk.org/gtk3/method.Widget.get_action_group.html)]

If GTK may not have provided any API for determining if a given action group is bound directly to the provided widget and/or bound directly to some containing widget, then I'm afraid it may be non-trivial to work around this limitation.

Secondly, I'm not certain if or how this could be applied for GTK 4 support. Having developed what seems like a patch that works for GTK 3, I thought it may be helpful to send the patch along for contribution/review.

This patch may at least serve to ensure that each Gio::SimpleActionGroup and each Gio::SimpleAction will be marked before GC, _at least once_.

Under local testing, the patch seems to work out for preventing that any custom action groups and any encapsulated Gio::SimpleAction instances would disappear under GC

### DIFF
--- a/gio2/ext/gio2/rb-gio2.c
+++ b/gio2/ext/gio2/rb-gio2.c
@@ -22,6 +22,46 @@
 
 #define RG_TARGET_NAMESPACE rb_mGio
 
+static void
+rb_gio2_action_mark(gpointer object)
+{
+    GAction *action = G_ACTION(object);
+    rbgobj_gc_mark_instance(action);
+}
+
+static void
+rb_gio2_action_proxy_mark(gpointer object)
+{
+    gchar **names;
+    gchar *name;
+    GAction *action;
+
+    names = g_action_group_list_actions(G_ACTION_GROUP(object));
+    for (size_t n = 0;  names[n] != NULL; n++) {
+        name = names[n];
+        action = g_action_map_lookup_action(G_ACTION_MAP(object), name);
+        if (action != NULL) {
+            rb_gio2_action_mark(action);
+        }
+    }
+    g_strfreev(names);
+}
+
+static void
+rb_gio2_application_mark(gpointer object)
+{
+    rb_gio2_action_proxy_mark(object);
+    rbgobj_gc_mark_instance(G_APPLICATION(object));
+}
+
+static void
+rb_gio2_simple_action_group_mark(gpointer object)
+{
+    rb_gio2_action_proxy_mark(object);
+    rbgobj_gc_mark_instance(G_SIMPLE_ACTION_GROUP(object));
+}
+
+
 void
 Init_gio2 (void)
 {
@@ -31,4 +71,8 @@ Init_gio2 (void)
 
     rb_gio2_init_application(RG_TARGET_NAMESPACE);
     rb_gio2_init_pollable_source(RG_TARGET_NAMESPACE);
+
+    rbgobj_register_mark_func(G_TYPE_ACTION, rb_gio2_action_mark);
+    rbgobj_register_mark_func(G_TYPE_APPLICATION, rb_gio2_application_mark);
+    rbgobj_register_mark_func(G_TYPE_SIMPLE_ACTION_GROUP, rb_gio2_simple_action_group_mark);
 }
--- a/gio2/ext/gio2/rb-gio2.c
+++ b/gio2/ext/gio2/rb-gio2.c
@@ -50,15 +50,19 @@ rb_gio2_action_proxy_mark(gpointer object)
 static void
 rb_gio2_application_mark(gpointer object)
 {
-    rb_gio2_action_proxy_mark(object);
-    rbgobj_gc_mark_instance(G_APPLICATION(object));
+    GApplication *app = G_APPLICATION(object);
+    rbgobj_gc_mark_instance(app);
+    if (g_application_get_is_registered(app)) {
+        rb_gio2_action_proxy_mark(app);
+    }
 }
 
 static void
 rb_gio2_simple_action_group_mark(gpointer object)
 {
-    rb_gio2_action_proxy_mark(object);
-    rbgobj_gc_mark_instance(G_SIMPLE_ACTION_GROUP(object));
+    GSimpleActionGroup *group = G_SIMPLE_ACTION_GROUP(object);
+    rbgobj_gc_mark_instance(group);
+    rb_gio2_action_proxy_mark(group);
 }
 
 
--- a/gio2/test/test-action-map.rb
+++ b/gio2/test/test-action-map.rb
@@ -48,4 +48,32 @@ class TestActionMap < Test::Unit::TestCase
       assert_equal([action, "X"], args)
     end
   end
+
+  sub_test_case "#add_action / #lookup_action / #remove_action" do
+    def setup
+      super
+      @action_name = "test"
+      @action = Gio::SimpleAction.new(@action_name)
+    end
+
+    test "add action, lookup action" do
+      @map.add_action(@action)
+      out = @map.lookup_action(@action_name)
+      assert_equal(@action, out)
+    end
+
+    test "add action, remove action, lookup action" do
+      @map.add_action(@action)
+      @map.remove_action(@action_name)
+      out = @map.lookup_action(@action_name)
+      assert_nil(out)
+    end
+
+    test "add action, GC, lookup action" do
+      @map.add_action(@action)
+      GC.start
+      out = @map.lookup_action(@action_name)
+      assert_equal(@action, out)
+    end
+  end
 end
--- a/gio2/test/test-application.rb
+++ b/gio2/test/test-application.rb
@@ -1,0 +1,49 @@
+# Copyright (C) 2022 Ruby-GNOME2 Project Team
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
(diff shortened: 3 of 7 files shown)


spchamp:
'm seeing a peculiar error with IRB, when  testing an application. I'm not certain it's due to this patch. I'd not noticed it before developing this patch, however, so I wanted to mention it here.

I would like to guess that it may be related to the additional mark handling for Gio::Application added in the patch. 

The backtrace from the segfault begins as follows
~~~~
(console:88999): GLib-GIO-CRITICAL **: 19:33:23.208: g_application_list_actions: assertion 'application->priv->is_registered' failed
/opt/rbsource/rbthinkum/rbcommon/rblib/vendor/bundle/ruby/3.1/gems/reline-0.3.1/lib/reline/unicode.rb:135: [BUG] Segmentation fault at 0x0000000000000000
ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [amd64-freebsd13
~~~~

It's a recurring error, though not easily reproducible. I'll try to figure out if the patch may have any side effect with or towards registration of a Gio::Application.

Here, I had to start IRB under something equivalent to `bundle exec irb`. Subsequently, after calling `require` then creating an instance of the Gtk::Application subclass I'm using it then segfaulted. 

This was while IRB was running command-line completion for `GC.start`. When it segfaulted, `GC.start` had not yet been run from IRB.

I've not seen this error when running the application via a 'bin' script. It may only be occurring when I try to initialize the application separately from an IRB console. 

Though I had not noticed this segfault earlier, but it also might not be due to this patch. 

Incidentally, the Gtk::Application subclass I'm using is using `type_register` under a local subclass that also uses `type_register`. Perhaps it might be easier to debug than to try to explain that rationale, lol

spchamp:
I've added two more changesets for the patch branch

[347108c](https://github.com/ruby-gnome/ruby-gnome/pull/1490/commits/347108cf9f14c8040ae8ade43133d604d96a055c) should serve to address the message ` GLib-GIO-CRITICAL **: 19:33:23.208: g_application_list_actions: assertion 'application->priv->is_registered' failed`, there avoiding any reflection on actions of a GApplication when the GAppliation has not yet been registered

[0bdce50](https://github.com/ruby-gnome/ruby-gnome/pull/1490/commits/0bdce5059534b0d56ff8143556f5bfbb57b6f77d) adds some additional tests for Gio::Action support. This includes two tests that would each call `GC.start` after adding an action, then before trying to retrieve information about the added action. This changeset also folds two of the previous tests in this patch branch into one test.

gio2/ext/gio2/rb-gio2.c
```diff
@@ -22,6 +22,50 @@
 
 #define RG_TARGET_NAMESPACE rb_mGio
 
+static void
+rb_gio2_action_mark(gpointer object)
+{
+    GAction *action = G_ACTION(object);
+    rbgobj_gc_mark_instance(action);
+}
```
kou, line 30: Why do we need this?

gio2/ext/gio2/rb-gio2.c
```diff
@@ -22,6 +22,50 @@
...
+        }
+    }
+    g_strfreev(names);
+}
+
+static void
+rb_gio2_application_mark(gpointer object)
+{
+    GApplication *app = G_APPLICATION(object);
+    rbgobj_gc_mark_instance(app);
```
kou, line 54: We don't need this.

gio2/ext/gio2/rb-gio2.c
```diff
@@ -22,6 +22,50 @@
...
+    if (g_application_get_is_registered(app)) {
```
kou, line 55: Why do we need this check?

gio2/ext/gio2/rb-gio2.c
```diff
@@ -22,6 +22,50 @@
...
 
+static void
+rb_gio2_action_mark(gpointer object)
+{
+    GAction *action = G_ACTION(object);
+    rbgobj_gc_mark_instance(action);
+}
+
+static void
+rb_gio2_action_proxy_mark(gpointer object)
```
kou, line 33: Can we register this only to `G_TYPE_ACTION_GROUP`?

gio2/ext/gio2/rb-gio2.c
```diff
@@ -22,6 +22,50 @@
...
+{
+    gchar **names;
+    gchar *name;
+    GAction *action;
+
+    names = g_action_group_list_actions(G_ACTION_GROUP(object));
+    for (size_t n = 0;  names[n] != NULL; n++) {
```
kou, line 40:
```suggestion
    size_t i;
    for (i = 0; names[i]; i++) {
```

gio2/ext/gio2/rb-gio2.c
```diff
@@ -22,6 +22,50 @@
...
+        name = names[n];
```
kou, line 41:
```suggestion
        gchar *name = names[n];
```

gio2/ext/gio2/rb-gio2.c
```diff
@@ -22,6 +22,50 @@
...
+    GAction *action = G_ACTION(object);
+    rbgobj_gc_mark_instance(action);
+}
+
+static void
+rb_gio2_action_proxy_mark(gpointer object)
+{
+    gchar **names;
+    gchar *name;
+    GAction *action;
```
kou, line 37:
```suggestion
```

gio2/ext/gio2/rb-gio2.c
```diff
@@ -22,6 +22,50 @@
...
+rb_gio2_action_mark(gpointer object)
+{
+    GAction *action = G_ACTION(object);
+    rbgobj_gc_mark_instance(action);
+}
+
+static void
+rb_gio2_action_proxy_mark(gpointer object)
+{
+    gchar **names;
```
kou, line 35:
```suggestion
    if (!G_IS_ACTION_MAP(object)) {
        return;
    }

    gchar **names;
```

gio2/ext/gio2/rb-gio2.c
```diff
@@ -22,6 +22,50 @@
...
+    gchar *name;
+    GAction *action;
+
+    names = g_action_group_list_actions(G_ACTION_GROUP(object));
+    for (size_t n = 0;  names[n] != NULL; n++) {
+        name = names[n];
+        action = g_action_map_lookup_action(G_ACTION_MAP(object), name);
```
kou, line 42:
```suggestion
        GAction *action = g_action_map_lookup_action(G_ACTION_MAP(object), name);
```

gio2/ext/gio2/rb-gio2.c
```diff
@@ -22,6 +22,50 @@
...
+        if (action != NULL) {
+            rb_gio2_action_mark(action);
+        }
```
kou, line 45:
```suggestion
        rbgobj_gc_mark_instance(action);
```

gio2/ext/gio2/rb-gio2.c
```diff
@@ -31,4 +75,8 @@ Init_gio2 (void)
 
     rb_gio2_init_application(RG_TARGET_NAMESPACE);
     rb_gio2_init_pollable_source(RG_TARGET_NAMESPACE);
+
+    rbgobj_register_mark_func(G_TYPE_ACTION, rb_gio2_action_mark);
+    rbgobj_register_mark_func(G_TYPE_APPLICATION, rb_gio2_application_mark);
+    rbgobj_register_mark_func(G_TYPE_SIMPLE_ACTION_GROUP, rb_gio2_simple_action_group_mark);
```
kou, line 81: We don't need them because we care this in https://github.com/ruby-gnome/ruby-gnome/blob/master/gio2/lib/gio2/action-map.rb .

kou:
I've fixed this with another approach: cefe9cb

spchamp:
Apologies about my delayed reply. I've had some personal business to attend to, which has been drawing the most of my attention. I've begun attending courses but will try to pick up on the API support, presently

PR #1491, opened 2022-07-28
glib2: print/parse methods for GVariant

spchamp:
This pull request would add a `GVariant.parse` class method and a `GVariant#variant_print` instance method.

Quoting the changelong:

rbglib-variant.c
- adding an instance method variant_print to GLib::Varaint
- adding a class method GLib::Variant.parse => Glib::Variant

glib2.rb
- adding a dependency for the added file glib2/variant.rb

glib2/variant.rb
- New file
- Renaming Variant.parse => Variant.parse_raw at runtime
- Dispatching to Variant.parse_raw from Variant.parse

test-variant.rb
- Adding tests for .parse, #variant_print methods in GLib::Variant

----

Known Limitations
- This initial change set does not provide any testing for the optional 'type' arg to the final `Variant.parse` method
- The definition for `Variant#print` is incomplete, in this changeset. The optional `type_annotate` arg is defined in the C source code, but then the arg is promptly lost due to the definition of the method interface for Ruby. As such, there's also no testing defined for this generally optional arg.

I'll try to address these in a subsequent change set. Health, all.

### DIFF
--- a/glib2/ext/glib2/rbglib-variant.c
+++ b/glib2/ext/glib2/rbglib-variant.c
@@ -330,6 +330,6 @@ Init_glib_variant(void)
     RG_DEF_METHOD(initialize, -1);
     RG_DEF_METHOD(value, 0);
     RG_DEF_METHOD(type, 0);
-    RG_DEF_METHOD(variant_print, 0);
+    RG_DEF_METHOD(variant_print, 1);
     RG_DEF_SMETHOD(parse, 2);
 }
--- a/glib2/ext/glib2/rbglib-variant.c
+++ b/glib2/ext/glib2/rbglib-variant.c
@@ -283,6 +283,43 @@ rg_type(VALUE self)
     return GVARIANTTYPE2RVAL((GVariantType *)g_variant_get_type(variant));
 }
 
+static VALUE
+rg_variant_print(VALUE self, VALUE type_annotate)
+{
+    GVariant *variant;
+    gboolean annotate;
+
+    variant = _SELF(self);
+    annotate = RVAL2CBOOL(type_annotate);
+
+    return CSTR2RVAL(g_variant_print(variant, annotate));
+}
+
+static VALUE
+rg_s_parse(VALUE self, VALUE str, VALUE rb_variant_type)
+{
+    const gchar *text = RVAL2CSTR(str);
+    const GVariantType *variant_type =
+        NIL_P(rb_variant_type) ? NULL : RVAL2GVARIANTTYPE(rb_variant_type);
+    const gchar *limit = NULL;
+    const gchar **endptr = NULL;
+    GVariant *variant;
+    GError *err = NULL;
+
+    variant = g_variant_parse(variant_type, text, limit, endptr, &err);
+
+    if (err != NULL) {
+        gchar* info = g_variant_parse_error_print_context(err, text);
+        g_log(G_LOG_DOMAIN, G_LOG_LEVEL_CRITICAL, info);
+        RAISE_GERROR(err);
+    }
+    VALUE rb_value = rg_variant_allocate(self);
+    g_variant_ref_sink(variant);
+    RTYPEDDATA_DATA(rb_value) = variant;
+    return rb_value;
+}
+
+
 void
 Init_glib_variant(void)
 {
@@ -293,4 +330,6 @@ Init_glib_variant(void)
     RG_DEF_METHOD(initialize, -1);
     RG_DEF_METHOD(value, 0);
     RG_DEF_METHOD(type, 0);
+    RG_DEF_METHOD(variant_print, 0);
+    RG_DEF_SMETHOD(parse, 2);
 }
--- a/glib2/lib/glib2/variant.rb
+++ b/glib2/lib/glib2/variant.rb
@@ -23,5 +23,11 @@ module GLib
         parse_raw(string, type)
       end
     end
+
+    alias_method :variant_print_raw, :variant_print
+
+    def variant_print(type_annotate = false)
+      variant_print_raw(type_annotate)
+    end
   end
 end
--- a/glib2/lib/glib2/variant.rb
+++ b/glib2/lib/glib2/variant.rb
@@ -1,0 +1,27 @@
+# Copyright (C) 2015-2022  Ruby-GNOME2 Project Team
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+module GLib
+  class Variant
+    class << self
+      alias_method :parse_raw, :parse
+
+      def parse(string, type = nil)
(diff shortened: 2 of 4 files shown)


spchamp:
I've added a second changeset for the patch branch, to address limitations of the initial changeset

Changelog:

GLib::Variant: Adding support for type_annotate in variant_print

rbglib-variant.c
- Using the correct arity for the variant_print method as defined in source

variant.rb
- Adding support for the type_annotate arg to variant_print as an optional arg

test-variant.rb
- Adding test for GLib::Variant.parse with optional Glib::VariantType
- Adding tests for GLib::Variant#variant_print with type_annotate arg

glib2/ext/glib2/rbglib-variant.c
```diff
@@ -283,6 +283,43 @@ rg_type(VALUE self)
...
+
+    return CSTR2RVAL(g_variant_print(variant, annotate));
+}
+
+static VALUE
+rg_s_parse(VALUE self, VALUE str, VALUE rb_variant_type)
+{
+    const gchar *text = RVAL2CSTR(str);
+    const GVariantType *variant_type =
+        NIL_P(rb_variant_type) ? NULL : RVAL2GVARIANTTYPE(rb_variant_type);
```
kou, line 303:
```suggestion
    const GVariantType *variant_type = RVAL2GVARIANTTYPE(rb_variant_type);
```

glib2/ext/glib2/rbglib-variant.c
```diff
@@ -283,6 +283,43 @@ rg_type(VALUE self)
...
+    const gchar *limit = NULL;
+    const gchar **endptr = NULL;
+    GVariant *variant;
+    GError *err = NULL;
+
+    variant = g_variant_parse(variant_type, text, limit, endptr, &err);
```
kou, line 309:
```suggestion
    GVariant *variant;
    GError *error = NULL;

    variant = g_variant_parse(variant_type, text, NULL, NULL, &error);
```

glib2/ext/glib2/rbglib-variant.c
```diff
@@ -283,6 +283,43 @@ rg_type(VALUE self)
...
+
+    if (err != NULL) {
+        gchar* info = g_variant_parse_error_print_context(err, text);
+        g_log(G_LOG_DOMAIN, G_LOG_LEVEL_CRITICAL, info);
+        RAISE_GERROR(err);
+    }
+    VALUE rb_value = rg_variant_allocate(self);
+    g_variant_ref_sink(variant);
+    RTYPEDDATA_DATA(rb_value) = variant;
+    return rb_value;
```
kou, line 319:
```suggestion
    return TypedData_Wrap_Struct(self, &rbg_variant_type, variant);
```

glib2/test/test-variant.rb
```diff
@@ -24,4 +24,71 @@ class TestGLibVariant < Test::Unit::TestCase
                    [variant.type, variant.value])
     end
   end
+
+  sub_test_case ".parse" do
+    test "value: true" do
```
kou, line 29: We don't need `value: ` prefix because this is not a keyword argument:

```suggestion
    test "true" do
```

glib2/test/test-variant.rb
```diff
@@ -24,4 +24,71 @@ class TestGLibVariant < Test::Unit::TestCase
...
+      variant = GLib::Variant.parse(Math::PI.to_s)
+      assert_equal(Math::PI, variant.value)
+    end
+
+    test "value: infinity" do
+      variant = GLib::Variant.parse("inf")
+      assert_equal(Float::INFINITY, variant.value)
+    end
+
+    test "value: int vector" do
```
kou, line 44:
```suggestion
    test "integer array" do
```

glib2/test/test-variant.rb
```diff
@@ -24,4 +24,71 @@ class TestGLibVariant < Test::Unit::TestCase
...
+      variant = GLib::Variant.parse("[1, 2, 3]")
+      assert_equal([1, 2, 3], variant.value)
+    end
+
+    test "value: integer, type: UINT16" do
```
kou, line 49:
```suggestion
    test "integer, UINT16" do
```

glib2/test/test-variant.rb
```diff
@@ -24,4 +24,71 @@ class TestGLibVariant < Test::Unit::TestCase
...
+      variant = GLib::Variant.parse("65535", GLib::VariantType::UINT16)
+      assert_equal(65535, variant.value)
+      assert_equal("q", variant.type.to_s)
```
kou, line 52: Could you use one assertion for related objects for easy to debug? If we use multiple assertions, the test is stopped when an assertion is failed. Rest assertions aren't executed and the result isn't logged.

```suggestion
      assert_equal([65535, "q"], [variant.value, variant.type.to_s])
```

glib2/test/test-variant.rb
```diff
@@ -24,4 +24,71 @@ class TestGLibVariant < Test::Unit::TestCase
...
+    test "value: true" do
+      variant = GLib::Variant.new(true)
+      out = variant.variant_print
+      assert_equal("true", out)
+    end
+
+    test "value: float" do
+      variant = GLib::Variant.new(Math::PI)
+      out = variant.variant_print
+      ## this may avoid mismatch from rounding error
```
kou, line 66: Can we remove this workaround by using other value such as `1.0`?

glib2/test/test-variant.rb
```diff
@@ -24,4 +24,71 @@ class TestGLibVariant < Test::Unit::TestCase
...
+
+    test "value: integer, type: UINT16" do
+      variant = GLib::Variant.parse("65535", GLib::VariantType::UINT16)
+      assert_equal(65535, variant.value)
+      assert_equal("q", variant.type.to_s)
+    end
+  end
+
+  sub_test_case "#variant_print" do
+    test "value: true" do
```
kou, line 57: We can remove `value: ` here because it's not a keyword argument:

 ```suggestion
     test "true" do
```

File: glib2/test/test-variant.rb [hunk 10]
```diff
@@ -24,4 +24,71 @@ class TestGLibVariant < Test::Unit::TestCase
...
+      assert_equal("inf", out)
+    end
+
+    test "value: int vector" do
+      variant = GLib::Variant.new([1, 2, 3])
+      out = variant.variant_print
+      assert_equal("[1, 2, 3]", out)
+    end
+
+    test "value: integer, type: UINT16, type_annotate: true" do
```
kou, line 82: We can remove `type: UINT16, ` because it's not important in this test:

```suggestion
    test "integer, type_annotate: true" do
```

glib2/lib/glib2/variant.rb
```diff
@@ -0,0 +1,33 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+module GLib
+  class Variant
+    class << self
+      alias_method :parse_raw, :parse
+
+      def parse(string, type = nil)
+        parse_raw(string, type)
+      end
```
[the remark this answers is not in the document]
spchamp: I'm not very familiar with the C API. I may be able to guess out how to implement an optional arg in C, from some existing example in the API.

Albeit from a limited perspective about available approaches, I'd thought it might be simpler to use a constant args list in C and mange the fully Ruby args syntax in Ruby?

glib2/ext/glib2/rbglib-variant.c
```diff
@@ -293,4 +330,6 @@ Init_glib_variant(void)
     RG_DEF_METHOD(initialize, -1);
     RG_DEF_METHOD(value, 0);
     RG_DEF_METHOD(type, 0);
+    RG_DEF_METHOD(variant_print, 1);
```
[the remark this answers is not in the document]
spchamp: Can do this. I'll try to add an annotation somewhere, about the API mapping.

Would it be well enough to leave the #inspect method as-is?

kou:
I've implemented `Variant#to_s`, `Variant#inspect` and `Variant.parse` to release a new version.

PR #1515, opened 2022-10-12
gtk4: fix builder.rb sample

kojix2:
* https://gitlab.gnome.org/GNOME/gtk/blob/main/examples/builder.c
* https://gitlab.gnome.org/GNOME/gtk/blob/main/examples/builder.ui


kou:
Thanks!

PR #1522, opened 2022-11-09
Always set wrap_struct_name for rbgobj

peterzhu2118:
Ruby requires all TypedData objects to have a wrap_struct_name. If it's null then ObjectSpace.dump_all will crash when it tries to read it.

### DIFF
--- a/glib2/ext/glib2/rbgobj_type.c
+++ b/glib2/ext/glib2/rbgobj_type.c
@@ -106,6 +106,7 @@ rbgobj_class_info_create_data_type(VALUE klass)
     rb_data_type_t *data_type;
 
     data_type = RB_ZALLOC(rb_data_type_t);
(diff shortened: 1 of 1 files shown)


glib2/ext/glib2/rbgobj_type.c
```diff
@@ -106,6 +106,7 @@ rbgobj_class_info_create_data_type(VALUE klass)
     rb_data_type_t *data_type;
 
     data_type = RB_ZALLOC(rb_data_type_t);
+    data_type->wrap_struct_name = "glib2/rbgobj";
```
kou, line 109:
```suggestion
    data_type->wrap_struct_name = "RGObjClassInfo";
```

kou (approved):
Thanks!

PR #1523, opened 2022-11-09
Fix memory leak in rbgobj

peterzhu2118:
TypedData objects requires the free function to also manage the allocated data buffer. Since rbgobj is always created with TypedData_Make_Struct, we have to also free the buffer to not leak memory.


peterzhu2118:
Looks like `call_cinfo_free` frees the memory later on, so this isn't needed.

kou:
You're right. We need to free it.
But we also need to unregister freed `cinfo` to avoid touching from others.

PR #1526, opened 2022-11-14
gtk4: Fix errors and deprecations in sample example files

viseztrance:
This fixes a few deprecation warnings and some errors in the drawing example. The `frame` has [been removed](https://gitlab.gnome.org/GNOME/gtk/-/commit/ece9e7e240bd91fc788343f0b6a9be3ccc52f81c) in the C example, so I removed it as well.

image


rubyFeedback:
Cool that this works - I was not aware that drawing widget works in ruby-gtk4 already. \o/

kou:
Thanks!

PR #1527, opened 2022-11-19
Add getting started instructions for GTK4

viseztrance:
This ports the [C getting started instructions](https://docs.gtk.org/gtk4/getting_started.html), up to the building applications section.


viseztrance (commented):
These should point to docs, but I can't find these anywhere on https://ruby-gnome.github.io/ruby-gnome/doc/dev/, so for now I'm pointing to their C apis.

gtk4/sample/examples/window_default.rb
```diff
@@ -0,0 +1,34 @@
...
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+# example from https://gitlab.gnome.org/GNOME/gtk/blob/main/examples/window-default.c
```
kou, line 19: Could you also add license information for the original code?

```suggestion
# example from https://gitlab.gnome.org/GNOME/gtk/blob/main/examples/window-default.c
#
# License: LGPL2.1+
```

gtk4/sample/getting-started/README.md
```diff
@@ -1,4 +1,357 @@
...
+  window.set_default_size(200, 200)
+  window.show
+end
+
+application.run
+```
+
+When creating a Gtk::Application you need to pick an application identifier (a name) and input to `Gtk::Application#new` as parameter. For this example *org.gtk.example* is used but for choosing an identifier for your application see this [guide](https://wiki.gnome.org/HowDoI/ChooseApplicationID).
+
+Lastly `Gtk::Application#new` takes a `Gio::ApplicationFlags` constant as input for your application, if your application would have special needs (those constants can be replaced by theirs respective symbol ie. `Gio::ApplicationFlags::FLAGS_NONE` == `:flags_none`). You must know that `GApplication` ignores arguments passed to `g_application_run()` on the Windows systems. It always uses command line arguments even when we pass an empty array to g_application_run().
```
kou, line 36:
```suggestion
Lastly `Gtk::Application#new` takes a `Gio::ApplicationFlags` constant as input for your application, if your application would have special needs (those constants can be replaced by theirs respective symbol ie. `Gio::ApplicationFlags::FLAGS_NONE` == `:flags_none`). You must know that `Gio:::Application` ignores arguments passed to `Gio::Application#run` on the Windows systems. It always uses command line arguments even when we pass an empty array to `Gio::Application#run`.
```

gtk4/sample/examples/window_default.rb
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env ruby
+#
+# Copyright (C) 2013-2022  Ruby-GNOME Project Team
```
kou, line 3:
```suggestion
# Copyright (C) 2022  Ruby-GNOME Project Team
```

kou (commented):
Thanks!

kou:
> These should point to docs, but I can't find these anywhere on https://ruby-gnome.github.io/ruby-gnome/doc/dev/, so for now I'm pointing to their C apis.

It makes sense.

PR #1528, opened 2022-12-02
rsvg2: Fix librsvg/rsvg-cairo.h header check

fweimer-rh:
The check always fails with current librsvg because the header cannot be included on its own anymore (despite the warning at the top).

Seen while working on:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC

### DIFF
--- a/rsvg2-no-gi/ext/rsvg2/extconf.rb
+++ b/rsvg2-no-gi/ext/rsvg2/extconf.rb
@@ -66,7 +66,7 @@ have_func("rsvg_handle_get_pixbuf_sub", rsvg_header)
 have_header("librsvg/rsvg-gz.h")
 have_type("RsvgDimensionData", "librsvg/rsvg.h")
 
-have_header("librsvg/rsvg-cairo.h")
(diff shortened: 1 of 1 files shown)


kou:
This code isn't used.
Could you show the error log you got?

kou:
I understand.

@mtasaka Why do you still maintain https://src.fedoraproject.org/rpms/ruby-gnome2 ? It seems that https://src.fedoraproject.org/rpms/rubygem-glib2 and so on can replace https://src.fedoraproject.org/rpms/ruby-gnome2 .

mtasaka:
rsvg2-no-gi is not used on Fedora, either. There is actually https://src.fedoraproject.org/rpms/rubygem-rsvg2 , this uses gi version.

The reason I keep old ruby-gnome2 is because some old packages depends on ruby-glade2 or ruby-gconf or so. which is no longer available from "new" ruby-gnome. But anyway even on Fedora rubygem-rsvg2 uses 4.0.3 gi version gem, so anyway this no-gi code (for rsvg2) is not used.

kou:
Thanks. I understand.

I close this.

PR #1552, opened 2023-03-01
gtk4: add sample/examples/application3

ToshioCP:
The gtk4/sample/examples/application3 directory is based on the C
programs in the following directory in the GTK 4 repository.

- https://gitlab.gnome.org/GNOME/gtk/-/blob/main/examples/application3

File opening codes and stack object are added to the `exampleapp.rb` in the aplication3 directory.


kou:
Thanks!

PR #1553, opened 2023-03-03
gtk4: Add sample/examples/application4

ToshioCP:
The original C program uses `gtk_action_map_add_action_entries` function.

- https://gitlab.gnome.org/GNOME/gtk/-/blob/main/examples/application4

But the program in this directory uses Ruby `each` iteration instead of the function above.
It makes the program simple and easy to understand.

### DIFF
--- a/gtk4/sample/examples/application4/exampleapp.rb
+++ b/gtk4/sample/examples/application4/exampleapp.rb
@@ -1,0 +1,147 @@
+#!/usr/bin/env ruby
+#
+# Copyright (C) 2023  Ruby-GNOME Project Team
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


gtk4/sample/examples/application4/exampleapp.rb
```diff
@@ -0,0 +1,147 @@
...
+
+    signal_connect "startup" do |application|
+      [ # action_name, handler (handler can be a Method object or Proc object)
+        ["preferences"],
+        ["quit", application.method(:quit)]
+      ].each do |action_name, handler|
+        action = Gio::SimpleAction.new(action_name)
+        action.signal_connect("activate") do |_action, _parameter|
+          handler.call if handler
+        end
```
kou, line 122: How about simplify this?

```suggestion
      [
        "preferences",
        "quit",
      ].each do |action_name,|
        action = Gio::SimpleAction.new(action_name)
        action.signal_connect("activate") do |_action, _parameter|
          __send__("#{action_name}_activated")
        end
```

And define `_activated` methods:

```ruby
private
def preferences_activated
end

def quit_activated
  quit
end
```

kou (commented):
> However, when we program GTK 4 with Ruby, add_action_entries method is not necessarily essential.
> Ruby has excellent features especially iterators.
> So, in my opinion, even if we can use add_action_entries, it is not the best way in Ruby language.

It makes sense.

ToshioCP:
I've update files along your advice.
And the pull request description is also updated.

kou:
Thanks!

PR #1554, opened 2023-03-05
gtk4: Add sample/examples/application5

ToshioCP:
Add gtk4/sample/examples/application5 diretory.

GSettints codes are added to `exampleapp.rb`.
It needs to compile the schema file before running the application.
But everything is done by rake so you don't need to do each task.
See `README.md` file for further information..


gtk4/sample/examples/application5/exampleapp.rb
```diff
@@ -0,0 +1,162 @@
...
+    view.cursor_visible = false
+    scrolled.child = view
+    stack.add_titled(scrolled, basename, basename)
+    buffer = view.buffer
+    stream = file.read
+    buffer.text = stream.read
+    tag = buffer.create_tag
+    @settings.bind("font", tag, "font", Gio::SettingsBindFlags::DEFAULT)
+    buffer.apply_tag(tag, buffer.start_iter, buffer.end_iter)
+    end
```
kou, line 113:
```suggestion
  end
```

kou (commented):
Hmm, how about setting `GSETTINGS_SCHEMA_DIR` in `exampleapp.rb` like https://github.com/ruby-gnome/ruby-gnome/blob/master/gtk3/sample/tutorial/exampleapp5/exampleapp.rb ?

Why do you prefer to this `install`/`uninstall` approach?

ToshioCP:
The program `gtk3/sample/tutorial/exampleapp5.rb` is simple in terms of all-in-one.
But I think it includes some redundant codes compared to applications as productions.
They are related to `glib-compile-schemas` (and `glib-compile-resources`).

On the other hand, the original C files are the same both in the developping and production phase.
It is because the differences, which are schema compilation and installation, are written in `meson.build`.
My idea is the same as it.

- The main Ruby program ` exampleapp.rb` is consistent in developping and production.
- The schema compilation and installation is controlled by another program, for example `install.rb` or `Rakefile`.

However, compiling the schema in `~/.local/share/glib-2.0/schemas` was carrying it too far, I think.
I should have done it in the current directory.

I won't stick to my idea.
All-in-one program is possible.

kou:
I see. Then I like the `Rakefile` approach:

```ruby
task default: :run

file "gschemas.compiled" => "org.gtk.exampleapp.gschema.xml" do
  # ...
end

task :run => "gschemas.compiled" do
  sh({"GSETTINGS_SCHEMA_DIR" => __dir__}, FileUtils::RUBY, "exampleapp.rb")
end

**Comment by ToshioCP** [comment]
Rakefile version.

I added two things to the Rakefile.

- Arguments can be given to `exampleapp.rb`
- Clean gschema.compiled

README.md is also updated.

File: gtk4/sample/examples/application5/README.md [hunk 2]
```diff
@@ -0,0 +1,39 @@
...
+
+```console
+$ cd gtk4/sample/examples/application5
+$ rake
+```
+
+If you want to give filenames to `exampleapp.rb`, use square bracket.
+
+```console
+$ rake default[README.md,Rakefile]
```
kou, line 27: Ah, we need to specify one ore more arguments to `exampleapp.rb`...

The Rake syntax isn't easy to use.
How about the followings?

1. We remove the `run` task and users run `exampleapp.rb` directly: `ruby exampeapp.rb README.md Rakefile`
2. `exampleapp.rb` sets `GSETTINGS_SCHEMA_DIR` internally.
3. `exampleapp.rb` checks `File.exist?(File.join(__dir__, "gschema.compiled"))` and reports an error when the check fails: `gschema.compiled doesn't exist. Run "rake" to generate tit.` or something

ToshioCP:
A new version.
The `Rakefile` only controls the schema file compilation.

I think it is simpler than the previous version, but not as simple as the all-in-one version.
If you prefer all-in-one version, I'll modify the files.
Feel free to give me your idea.

gtk4/sample/examples/application5/exampleapp.rb
```diff
@@ -0,0 +1,168 @@
...
+
+  def quit_activated
+    quit
+  end
+end
+
+if File.exist?(File.join(__dir__, "gschemas.compiled"))
+  ENV["GSETTINGS_SCHEMA_DIR"] = __dir__
+else
+ raise %{gschemas.compiled doesn't exist. Run "rake" to generate it.}
```
kou, line 159:
```suggestion
  raise %{gschemas.compiled doesn't exist. Run "rake" to generate it.}
```

gtk4/sample/examples/application5/Rakefile
```diff
@@ -0,0 +1,11 @@
+require 'rake/clean'
```
kou, line 1: Could you use double quote for consistency?

```suggestion
require "rake/clean"
```

kou (commented):
I'm OK with this approach.
Thanks.

PR #1556, opened 2023-03-10
gtk4: Add sample/examples/application7

ToshioCP:
Add gtk4/sample/examples/application7

### DIFF
--- a/gtk4/sample/examples/application7/exampleapp.rb
+++ b/gtk4/sample/examples/application7/exampleapp.rb
@@ -17,12 +17,12 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 #
 # Example from:
-# * https://gitlab.gnome.org/GNOME/gtk/-/blob/main/examples/application6/exampleapp.c
-# * https://gitlab.gnome.org/GNOME/gtk/-/blob/main/examples/application6/exampleappwin.c
-# * https://gitlab.gnome.org/GNOME/gtk/-/blob/main/examples/application6/exampleappprefs.c
-# * https://gitlab.gnome.org/GNOME/gtk/-/blob/main/examples/application6/window.ui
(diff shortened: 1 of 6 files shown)


gtk4/sample/examples/application7/exampleapp.rb
```diff
@@ -0,0 +1,307 @@
...
+          </item>
+        </section>
+      </menu>
+    </interface>
+    MENU
+    super(application: application)
+    builder = Gtk::Builder.new(string: menu_ui)
+    gears.menu_model = builder["menu"]
+    @settings = Gio::Settings.new("org.gtk.exampleapp")
+    @settings.bind("transition", stack, "transition-type", Gio::SettingsBindFlags::DEFAULT)
```
kou, line 215: Can we use `:default` here?

```suggestion
    @settings.bind("transition", stack, "transition-type", :default)
```

gtk4/sample/examples/application7/exampleapp.rb
```diff
@@ -0,0 +1,307 @@
...
+  private
+
+  def search_text_changed(search_entry)
+    text = search_entry.text
+    return if text.empty?
+
+    tab = stack.visible_child
+    view = tab.child
+    buffer = view.buffer
+    range = buffer.start_iter.forward_search(text, Gtk::TextSearchFlags::CASE_INSENSITIVE)
```
kou, line 247:
```suggestion
    range = buffer.start_iter.forward_search(text, :case_insensitive)
```

ToshioCP:
How about `Gio::SettingsBindFlags::DEFAULT`?
It is used in three places.

kou:
If we can use `:default`, I want to use `:default` in all places.

ToshioCP:
I've changed all Gio::SettingsBingFlags::DEFAULT to :default.

kou:
Did `:default` work?

ToshioCP:
Yes.
They worked at least in my environment (UBUNTU 22.10).

kou:
Thanks!

PR #1557, opened 2023-03-11
gtk4: Add sample/examples/application8

ToshioCP:
Add gtk4/sample/examples/application8

### DIFF
--- a/gtk4/sample/examples/application8/exampleapp.rb
+++ b/gtk4/sample/examples/application8/exampleapp.rb
@@ -1,0 +1,358 @@
+#!/usr/bin/env ruby
+#
+# Copyright (C) 2023  Ruby-GNOME Project Team
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# Example from:
+# * https://gitlab.gnome.org/GNOME/gtk/-/blob/main/examples/application8/exampleapp.c
+# * https://gitlab.gnome.org/GNOME/gtk/-/blob/main/examples/application8/exampleappwin.c
+# * https://gitlab.gnome.org/GNOME/gtk/-/blob/main/examples/application8/exampleappprefs.c
+# * https://gitlab.gnome.org/GNOME/gtk/-/blob/main/examples/application8/window.ui
+# * https://gitlab.gnome.org/GNOME/gtk/-/blob/main/examples/application8/gears-menu.ui
+# * https://gitlab.gnome.org/GNOME/gtk/-/blob/main/examples/application8/prefs.ui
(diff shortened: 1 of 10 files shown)


ToshioCP:
The script `exampleapp.rb` uses ruby methods instead of C functions.

```ruby
291:     strings = buffer.text.downcase.split(/\W+/).uniq
292:     strings.delete("")
```

It can be modified to use C functions like this:

```ruby
    strings = []
    start_iter = buffer.start_iter
    until start_iter.end?
      until start_iter.starts_word || start_iter.end?
        start_iter.forward_char
      end
      break if start_iter.end?
      end_iter = start_iter.clone
      end_iter.forward_word_end
      word = buffer.get_text(start_iter, end_iter, false).downcase
      strings << word
      start_iter = end_iter
    end
    strings = strings.uniq
```

If you prefer the code above, I'll modify it.

gtk4/sample/examples/application8/exampleapp.rb
```diff
@@ -0,0 +1,358 @@
...
+    @settings = Gio::Settings.new("org.gtk.exampleapp")
+    @settings.bind("transition", stack, "transition-type", :default)
+    @settings.bind("show-words", sidebar, "reveal-child", :default)
+    search.bind_property("active", searchbar, "search-mode-enabled", :bidirectional)
+    sidebar.signal_connect "notify::reveal-child" do |_sidebar, _pspec|
+      update_words
+    end
+    action = @settings.create_action("show-words")
+    add_action(action)
+end
```
kou, line 249:
```suggestion
  end
```

gtk4/sample/examples/application8/exampleapp.rb
```diff
@@ -0,0 +1,358 @@
...
+    buffer.select_range(range[0], range[1])
+    view.scroll_to_iter(range[0], 0.0, false, 0.0, 0.0)
+  end
+
+  def update_words
+    tab = stack.visible_child
+    return unless tab
+    view = tab.child
+    buffer = view.buffer
+    strings = buffer.text.downcase.split(/\W+/).uniq
```
kou, line 291: `\W+` doesn't work with non-ASCII characters such as Japanese word. How about using the default separator?

```suggestion
    strings = buffer.text.downcase.split.uniq
```

kou (commented):
Let's use Ruby implementation.
Ruby implementation is easy to read and calling many C functions is slow.

gtk4/sample/examples/application8/exampleapp.rb
ToshioCP: The original program picks up "words" which are defined in Pango.
And Pango supports Unicode.
Words don't include ";",  ":" or some other characters.
The default argument of the split method is `$;` and `$;` is usually nil.
Then split uses space characters as delimiters, which don't include ";" and ":".
Therefore, the behavior is different.
For example,

Sample text:
abc def; ghi.

Original program:
=> ["abc", "def", "ghi"]

Split method with the default argument:
=> ["abc", "def;", "ghi."]

I think maybe the following code works in Unicode texts.

```ruby
strings = buffer.text.downcase.split(/[^[:word:]]+/).uniq
```

How about it?

ToshioCP: I found that `example.app` didn't support Japanese text.
Maybe we need to modify:

`stream.read` => `stream.read.force_encoding(Encoding::UTF_8)`

kou: `strings = buffer.text.downcase.scan(/[[:word:]]+/).uniq` is better. We can remove `strings.delete("")` with it.

```ruby
file.read do |stream|
  buffer.text = stream.read.force_encoding(Encoding::UTF_8)
end
```

is better.

ToshioCP:
I've changed exampleapp.rb to support UTF-8 encoding.
In addition,aplication3/exampleapp.rb to application7/exampleapp.rb are also fixed.

kou:
Thanks!

Issue #1611, opened 2024-03-30
Many broken links to OSDN


znz:
There are many links to OSDN in README, hompage in gemspec, and website URL of this repository details.
But it is not accessible.

So I have a problem to reach to this repository from https://rubygems.org/gems/glib2 and other ruby-gnome2 gems at rubygems.org.

related: https://github.com/ruby-gnome/ruby-gnome/issues/1465

kou:
Could you open a PR to update URLs?

znz:
I don't know where the primary website is.
And I can't change website URL of this repository details.

kou:
https://ruby-gnome.github.io/

You don't need to do everything. You can do what you can do.

I've updated the repository details.

PR #1615, opened 2024-05-11
Fix broken links

znz:
Replace "https://ruby-gnome2.osdn.jp/" with "https://ruby-gnome.github.io/" in homepage and Project Website. 

Fix https://github.com/ruby-gnome/ruby-gnome/issues/1611


kou:
Thanks.

PR #1643, opened 2024-12-22
Create list_box_demo.rb

Beagle123:
I like this control.  Its really simple.  It was refreshing to have learning a control take 10 minutes instead of a day or two.

I think this one is ready to go without issues.

### DIFF
--- a/gtk4/sample/examples/list_box_demo.rb
+++ b/gtk4/sample/examples/list_box_demo.rb
@@ -1,0 +1,64 @@
+#!/usr/bin/env ruby
+#
+# Copyright (C) 2024  Ruby-GNOME Project Team
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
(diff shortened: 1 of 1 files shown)


gtk4/sample/examples/list_box_demo.rb
```diff
@@ -0,0 +1,64 @@
...
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# Example from:
+# * https://gitlab.gnome.org/GNOME/gtk/-/blob/main/examples/application1/exampleapp.c
+# * https://gitlab.gnome.org/GNOME/gtk/-/blob/main/examples/application1/exampleappwin.c
+# License: LGPL2.1-or-later
```
kou, line 22: It this really based on these files?

Beagle123: No,  you can delete those lines

Issue #1654, opened 2025-01-07
glib2: support `TypeInstance` in `Value` conversions


ppibburr:
noticed when an interface implements another, and then a class implements those, properties get wrapped in values
Currently there are no value conversions anyways?.

in `rbgobj_value.c`

I made these two changes

`rbgobj_rvalue_to_gvalue`
```c
default:
        if (!rbgobj_convert_rvalue2gvalue(fundamental_type, val, result)) {
            RValueToGValueFunc func =
                g_type_get_qdata(type, qRValueToGValueFunc);
            if (!func){
                g_value_set_instance(result, rbgobj_instantiatable_get(val));
				
                //g_warning("rbgobj_rvalue_to_gvalue: unsupported type: %s\n",
                //          g_type_name(type));
            } else {
```

and

`rbgobj_gvalue_to_rvalue`
```c
            if (!func) {
		GType type = G_TYPE_FROM_INSTANCE(g_value_peek_pointer(value));
		if (GTYPE2CLASS(type)!=NULL) {
			VALUE rb_instance = rbgobj_instantiatable_to_ruby(g_value_peek_pointer(value), TRUE);
			rbgobj_instantiatable_initialize(rb_instance, g_value_peek_pointer(value));
			return rb_instance;
		}
				
                g_warning("rbgobj_gvalue_to_rvalue: unsupported type: %s\n",
                          g_type_name(type));
```

makes this case work

Rakefile follows: I broke `set` and `get` into task
`rake set`
`rake get`
to test independently

```ruby
VER = '0.1'

directory "./build"
directory "./src"

file "./src/bar.vala" => ["./src"] do |t|
  File.open(t.name, "w") do |f|
    f.puts <<~EOC
      namespace Foo {
        public class Moof {
          public Moof me() {
            return this;
          }
        }
        public interface IFace : Object { 
          public abstract Moof? other {get; set;}
        }
        public class Bar : IFace, Object { 
          construct {
            this.other = new Moof();
          }
          public Moof? other {get; set;}

        }
      }
    EOC
  end
end

file "foo.rb" do |t|
  File.open(t.name, "w") do |f|
    f.puts <<~EOC
      require 'gobject-introspection'

      module Foo
        class Loader < GObjectIntrospection::Loader
        end

        Loader.load("Foo", self)
      end
      
      # Fails get
      def get
        p :get
        b=Foo::Bar.new()
      
        b.other
      end
      
      # Fails set
      def set
        p :set
[...]

PR #1653, opened 2025-01-06
glib2: add `GLib::TypeInstance#inspect`

ppibburr:
GitHub: fix GH-1652

PR #1658, opened 2025-01-10
glib2: implement `Variant#inspect`

ppibburr:
fix tests of GH-1653


glib2/lib/glib2/variant.rb
ppibburr, line 1: fix GH-1654

ppibburr:
GH-1653  along with this should pass all tests now from `main` branch

Issue #1657, opened 2025-01-09
gobject-introspection: `rb_gi_arguments_convert_return_value_free_everything_interface` doesn't use `ref` and `unref` from type_info for GI_INFO_TYPE_OBJECT


ppibburr:
results in warnings to stderr and not actually calling `unref` on the instance

this seems to make it work: in `rb_gi_arguments.c`
```c
      case GI_INFO_TYPE_OBJECT:
        {
						GIObjectInfoUnrefFunction unref = g_object_info_get_unref_function_pointer(data->metadata->type.interface_info);
            GObject *object = data->value->v_pointer;
            if (G_IS_OBJECT(object)) {
								if (g_object_is_floating(object)) {
										g_object_ref_sink(object);
								}
						}
            unref(object);
```

PR #1659, opened 2025-01-11
gi: use `ref`/`unref` for `GTypeInstance`

ppibburr:
fix GH-1657

### DIFF
--- a/gobject-introspection/ext/gobject-introspection/rb-gi-arguments.c
+++ b/gobject-introspection/ext/gobject-introspection/rb-gi-arguments.c
@@ -2156,7 +2156,7 @@ rb_gi_arguments_convert_return_value_free_everything_interface(
             GObject *object = data->value->v_pointer;
             if (G_IS_OBJECT(object)) {
                 if (g_object_is_floating(object)) {
-                        g_object_ref_sink(object);
+                    g_object_ref_sink(object);
                 }
                 g_object_unref(object);
             } else {
--- a/gobject-introspection/ext/gobject-introspection/rb-gi-arguments.c
+++ b/gobject-introspection/ext/gobject-introspection/rb-gi-arguments.c
@@ -604,7 +604,7 @@ rb_gi_arguments_convert_arg_interface_body(VALUE user_data)
       case GI_INFO_TYPE_INTERFACE:
         if (!G_IS_OBJECT(data->arg->v_pointer)) {
             GIObjectInfoRefFunction ref = g_object_info_get_ref_function_pointer(data->arg_metadata->type.interface_info);
-            if (ref != NULL) {
+            if (ref) {
                 ref(data->arg->v_pointer);
             }
         }
--- a/gobject-introspection/ext/gobject-introspection/rb-gi-arguments.c
+++ b/gobject-introspection/ext/gobject-introspection/rb-gi-arguments.c
@@ -2147,11 +2147,16 @@ rb_gi_arguments_convert_return_value_free_everything_interface(
         break;
       case GI_INFO_TYPE_OBJECT:
         {
+            GIObjectInfoUnrefFunction unref = g_object_info_get_unref_function_pointer(data->metadata->type.interface_info);
             GObject *object = data->value->v_pointer;
-            if (g_object_is_floating(object)) {
-                g_object_ref_sink(object);
+            if (G_IS_OBJECT(object)) {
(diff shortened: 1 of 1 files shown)


gobject-introspection/ext/gobject-introspection/rb-gi-arguments.c
```diff
@@ -2147,11 +2147,16 @@ rb_gi_arguments_convert_return_value_free_everything_interface(
         break;
       case GI_INFO_TYPE_OBJECT:
         {
+            GIObjectInfoUnrefFunction unref = g_object_info_get_unref_function_pointer(data->metadata->type.interface_info);
```
kou, line 2150: Could you move this to the `else` clause?

ppibburr:
seems that `TypeInstance` does not get the `refcount` incremented as `GObject`s do in `rbgobj_get_ruby_object_from_gobject`?

This gives one less refcount versus Object at times, which then this PR will unref to `0` when the refcount should still `>0`

```c
    } else if (alloc) {
        VALUE obj;

        obj = rbgobj_object_alloc_func(GTYPE2CLASS(G_OBJECT_TYPE(gobj)));
        gobj = g_object_ref(gobj);
```

```c
    if (alloc) {
        GType type = G_TYPE_FROM_INSTANCE(instance);
        VALUE rb_instance = instantiatable_s_allocate(GTYPE2CLASS(type));
        
        rbgobj_instantiatable_initialize(rb_instance, instance);
        return rb_instance;
```

kou:
It seems that we need to `ref()` here: https://github.com/ruby-gnome/ruby-gnome/blob/d7a4ce831cc2ea90da64cce6baee386c26d0e359/gobject-introspection/ext/gobject-introspection/rb-gi-arguments.c#L605

ppibburr:
would say this good to go...

Wondering about implementing `GLib::Instantiable#unref`
Also in `rbg_glib_instantiatable_holder_free`, having it call the unref function for the GITypeInfo

gobject-introspection/ext/gobject-introspection/rb-gi-arguments.c
```diff
@@ -602,6 +602,12 @@ rb_gi_arguments_convert_arg_interface_body(VALUE user_data)
         }
       case GI_INFO_TYPE_OBJECT:
       case GI_INFO_TYPE_INTERFACE:
+        if (!G_IS_OBJECT(data->arg->v_pointer)) {
+            GIObjectInfoRefFunction ref = g_object_info_get_ref_function_pointer(data->arg_metadata->type.interface_info);
+            if (ref != NULL) {
```
kou, line 607:
```suggestion
            if (ref) {
```

gobject-introspection/ext/gobject-introspection/rb-gi-arguments.c
```diff
@@ -2148,10 +2154,17 @@ rb_gi_arguments_convert_return_value_free_everything_interface(
...
-            if (g_object_is_floating(object)) {
-                g_object_ref_sink(object);
+            if (G_IS_OBJECT(object)) {
+                if (g_object_is_floating(object)) {
+                        g_object_ref_sink(object);
+                }
+                g_object_unref(object);
+            } else {
+                GIObjectInfoUnrefFunction unref = g_object_info_get_unref_function_pointer(data->metadata->type.interface_info);
+                if (unref != NULL) {
```
kou, line 2164:
```suggestion
                if (unref) {
```

gobject-introspection/ext/gobject-introspection/rb-gi-arguments.c
```diff
@@ -2148,10 +2154,17 @@ rb_gi_arguments_convert_return_value_free_everything_interface(
       case GI_INFO_TYPE_OBJECT:
         {
             GObject *object = data->value->v_pointer;
-            if (g_object_is_floating(object)) {
-                g_object_ref_sink(object);
+            if (G_IS_OBJECT(object)) {
+                if (g_object_is_floating(object)) {
+                        g_object_ref_sink(object);
```
kou, line 2159:
```suggestion
                    g_object_ref_sink(object);
```

kou (commented):
We can move this to glib2 later because girepository API is moved to GLib from GObject Introspection since GLib 2.80.0.

kou:
Thanks.

PR #1687, opened 2025-09-26
glib2: Add support for param for `TypeInstance` based object

ppibburr:
Vala uses `TypeInstance` when no parent isn't defined:

```vala
public class MyClass { // TypeInstance is used
}
```

`GLib::Param::Object.new` uses `g_param_spec_object()` internally and it validates whether the given `GType` is a `G_TYPE_OBJECT`. But `TypeInstance` based object isn't a `G_TYPE_OBJECT`. So we can't use `g_param_spec_object()` for `TypeInstance` based object that is created by Vala.

Vala uses `g_param_spec_internal(G_TYPE_PARAM_OBJECT)` for `TypeInstance` based object. So we add `GLib::Param::TypeInstance` for the case.

### DIFF
--- a/glib2/ext/glib2/rbgobj_paramspecs.c
+++ b/glib2/ext/glib2/rbgobj_paramspecs.c
@@ -264,6 +264,21 @@ object_initialize(VALUE self, VALUE name, VALUE nick, VALUE blurb,
                   VALUE object_type, VALUE flags)
 {
     GParamSpec* pspec;
+    pspec = g_param_spec_object(RVAL2CSTR(name),
+                                RVAL2CSTR_ACCEPT_NIL(nick),
+                                RVAL2CSTR_ACCEPT_NIL(blurb),
+                                rbgobj_gtype_from_ruby(object_type),
+                                resolve_flags(flags));
+    
+    rbgobj_param_spec_initialize(self, pspec);
+    return Qnil;
+}
+
+static VALUE
+typeinstance_initialize(VALUE self, VALUE name, VALUE nick, VALUE blurb,
+                  VALUE object_type, VALUE flags)
+{
+    GParamSpec* pspec;
     pspec = g_param_spec_internal(G_TYPE_PARAM_OBJECT,
                                 RVAL2CSTR(name),
                                 RVAL2CSTR_ACCEPT_NIL(nick),
@@ -356,4 +371,7 @@ Init_gobject_gparamspecs(void)
 
     c = G_DEF_CLASS(G_TYPE_PARAM_OBJECT, "Object", cParamSpec);
     rb_define_method(c, "initialize", object_initialize, 5);
+    
+    c = G_DEF_CLASS(G_TYPE_PARAM_OBJECT, "TypeInstance", cParamSpec);
+    rb_define_method(c, "initialize", typeinstance_initialize, 5);
 }


kou:
Could you show a problem case?

ppibburr:
```vala
namespace Ree {
  public class Bar { // TypeInstance NOT Object
  }
 
  // include Foo
  // install_property(GLib::Param::Object.new('bar','bar','bar', Ree::Foo,GLib::Param::READABLE, GLib::Param::WRITABLE)) 
  // # fails because +Bar+ is not an +Object+
  public interface Foo {
    public abstract Bar bar {get;set;}
  }
}
```
creates this C code

```c
GParamSpec*
ree_param_spec_bar (const gchar* name,
                    const gchar* nick,
                    const gchar* blurb,
                    GType object_type,
                    GParamFlags flags)
{
	ReeParamSpecBar* spec;
	g_return_val_if_fail (g_type_is_a (object_type, REE_TYPE_BAR), NULL);
	spec = g_param_spec_internal (G_TYPE_PARAM_OBJECT, name, nick, blurb, flags);
	G_PARAM_SPEC (spec)->value_type = object_type;
	return G_PARAM_SPEC (spec);
}
```

```c
static void
ree_foo_default_init (ReeFooIface * iface,
                      gpointer iface_data)
{
	g_object_interface_install_property (iface, ree_param_spec_bar ("bar", "bar", "bar", REE_TYPE_BAR, G_PARAM_STATIC_STRINGS | G_PARAM_READABLE | G_PARAM_WRITABLE));
}
```

ppibburr:
When implementing an interface in ruby, properties derived from TypeInstance and not Object, vala just treats them as an Object param type with a custom constructor

....

Currently, this patch fixes this issue
As another issue, but I've not traced it down to the cause, a ruby program will segfault before the ruby class that implements the interface's code `get/set_property_func` gets called ...

So for now I do this:
```vala
public interface Foo : Object {
  public Bar bar { 
    get {
      var v=Value(typeof(Bar));
      this.get_property("bar", ref v);
      return ((Bar)v);
    }

    set {
      var v=Value(typeof(Bar));
      v = value;
     this.set_property("bar", v);
   }
}
```
and it eliminates the need for this patch. Not sure if the segfault can be avoided with another patch or is a Vala improvement

ppibburr:
```ruby
VALA = <<~EOV
	namespace Ree {
		public interface Foo : Object {

			public Bar bar {
			  get {
				var v=Value(typeof(Bar));
				this.get_property("bar", ref v);
				return ((Bar)v);
			  }
			  
			  set {
				var v=Value(typeof(Bar));
				v = value;
				this.set_property("bar", v);
			  }
			}
			
			public void test() {
				bar.woof();
			}
		}
		
		public class Bar {
			public void woof() {
				stdout.printf("woof\n");
			}
		}
		
		public class Quux : Foo, Object {
			public Bar bar {get;set;}
		}
		
		static void main() {
			stdout.printf("e\n");
		}
	}
EOV

RUBY = <<~EOR
	require 'gobject-introspection'

	module Ree
	  class Loader < GObjectIntrospection::Loader
	  end

	  Loader.load("Ree", self)
	end


	class Moof < GLib::Object
	  type_register
	  
	  include Ree::Foo
	  
	  install_property(GLib::Param::Object.new(
		"bar",
		"bar",
		"bar",
		Ree::Bar,
		GLib::Param::READABLE | GLib::Param::WRITABLE
	  ))

	  def bar
		@bar
	  end
	  
	  def bar= b
		@bar=b
	  end
	end

	p ree=Moof.new
	p ree.set_property('bar',Ree::Bar.new)
	p ree.get_property('bar')
	ree.test
EOR

task :clean do
  `mkdir -p ./ppibburr_sample`
  `rm ./ppibbur_sample/*.c ./ppibburr_sample/*.o ./ppibburr_sample/*.gir ./ppibburr_sample/*.h ./ppibburr_sample/*.so ./ppibburr_sample/*.typelib`
end

task :default => :clean do
  `mkdir -p ./ppibburr_sample`
  
  File.open("./ppibburr_sample/ree.vala", "w") do |o|
    o.puts VALA
  end
  
  File.open("./ppibburr_sample/ree.rb", "w") do |o|
    o.puts RUBY
  end
  
  Dir.chdir "./ppibburr_sample"
  
  system "valac ~/ree.vala -o libree.so -X -fPIC -X -shared -H ree.h --library Ree-1.0 --gir Ree-1.0.gir"
  system "g-ir-compiler --shared-library=./libree.so -o Ree-1.0.typelib Ree-1.0.gir"
  system "LD_LIBRARY_PATH=./ GI_TYPELIB_PATH=./ ruby ./ree.rb"
end
```
as a Rakefile to test

ppibburr:
`GLib::Param::TypeInstance` is perfect

glib2/ext/glib2/rbgobj_paramspecs.c
```diff
@@ -269,6 +269,23 @@ object_initialize(VALUE self, VALUE name, VALUE nick, VALUE blurb,
                                 RVAL2CSTR_ACCEPT_NIL(blurb),
                                 rbgobj_gtype_from_ruby(object_type),
                                 resolve_flags(flags));
+
```
kou, line 272: Could you revert needless change?

```suggestion
```

glib2/ext/glib2/rbgobj_paramspecs.c
```diff
@@ -269,6 +269,23 @@ object_initialize(VALUE self, VALUE name, VALUE nick, VALUE blurb,
...
+typeinstance_initialize(VALUE self, VALUE name, VALUE nick, VALUE blurb,
+                  VALUE object_type, VALUE flags)
+{
+    GParamSpec* pspec;
+    pspec = g_param_spec_internal(G_TYPE_PARAM_OBJECT,
+                                RVAL2CSTR(name),
+                                RVAL2CSTR_ACCEPT_NIL(nick),
+                                RVAL2CSTR_ACCEPT_NIL(blurb),
+                                resolve_flags(flags));
+
```
kou, line 287:
```suggestion
```

glib2/ext/glib2/rbgobj_paramspecs.c
```diff
@@ -269,6 +269,23 @@ object_initialize(VALUE self, VALUE name, VALUE nick, VALUE blurb,
...
+static VALUE
+typeinstance_initialize(VALUE self, VALUE name, VALUE nick, VALUE blurb,
+                  VALUE object_type, VALUE flags)
+{
+    GParamSpec* pspec;
+    pspec = g_param_spec_internal(G_TYPE_PARAM_OBJECT,
+                                RVAL2CSTR(name),
+                                RVAL2CSTR_ACCEPT_NIL(nick),
+                                RVAL2CSTR_ACCEPT_NIL(blurb),
+                                resolve_flags(flags));
```
kou, line 286:
```suggestion
                                  RVAL2CSTR(name),
                                  RVAL2CSTR_ACCEPT_NIL(nick),
                                  RVAL2CSTR_ACCEPT_NIL(blurb),
                                  resolve_flags(flags));
```

PR #1706, opened 2026-04-09
Fix documentation generation task

KitaitiMakoto:
Hello,

I found `doc:generate` task fails and fixed it.

Thank you.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -30,7 +30,7 @@ group :development, :docs, :test do
 end
 
 group :docs do
-  gem "commonmarker"
+  gem "commonmarker", "~> 0.21"
   gem "rdoc"
   gem "yard"
   gem "yard-gobject-introspection",
--- a/gvlc/.yardopts
+++ b/gvlc/.yardopts
@@ -1,0 +1,23 @@
+# Copyright (C) 2026  Ruby-GNOME Project Team
+#
+# This library is free software; you can redistribute it and/or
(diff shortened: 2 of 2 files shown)


kou:
Good catch!

Gemfile
```diff
@@ -30,7 +30,7 @@ group :development, :docs, :test do
 end
 
 group :docs do
-  gem "commonmarker"
+  gem "commonmarker", "~> 0.21"
```
kou, line 33: Hmm. Why do we need to use old commonmarker?
YARD doesn't do it: https://github.com/lsegal/yard/blob/1e6d517e228e73812410f65f9a6d8fcb91195cd3/Gemfile#L21

It seems that there are some known vulnerabilities for old commonmarker...

KitaitiMakoto: The reason why is that YARD doesn't follow the class name change of common marker: https://github.com/lsegal/yard/issues/1528

Tests for commonmarker is skipped now in YARD repository.

So, I should have sent a pull request to YARD rather than this project. BTW, Thank you for merge.

kou: Ah, I understand.

Let's fix it in YARD.

BTW, we may migrate to RDoc because I'm one of maintainers of RDoc recently.

KitaitiMakoto: There are already pull requests:

* https://github.com/lsegal/yard/pull/1540
* https://github.com/lsegal/yard/pull/1601

 I keep an eye on them.

PR #1704, opened 2026-04-02
gstreamer: Make audio Gst::Sample export MemoryView

KitaitiMakoto:
Hi,

I made `Gst::Sample` export MemoryView for #1634.
(Sorry for big commits.)

But, it's not perfect. Could you advise me?

* Only audio is supported, because my primary motivation is speech-to-text from microphone. Is video support needed?
* Only row-major data is supported
* Only interleaved audio is supported
* Some formats like `GST_AUDIO_FORMAT_S24_32LE` are not supported because I couldn't find way to handle them. Is there a way to export such format to MemoryView?
* I'm not really sure of endianness because I'm not familiar with low level programming
* I'm not sure of the way of setup sample audio data in test is good or not. Current way to setup audio (`generate_samples`) is not performant

Any advice is appreciated.

Example for speech-to-text from microphone below works with this pull request and whispercpp's [pre-pull-request branch](https://github.com/KitaitiMakoto/whisper.cpp/):

```ruby:sample/asr.rb
require "gst"
require "whisper"

pipeline = Gst::Pipeline.new("asr")
src = Gst::ElementFactory.make("autoaudiosrc", nil)
raise "need audiotestsrc from gst-plugins-good" if src.nil?
resample = Gst::ElementFactory.make("audioresample", nil)
raise "need audioresample from gst-plugins-base" if resample.nil?
sink = Gst::ElementFactory.make("appsink", nil)
raise "need appsink from gst-plugins-base" if sink.nil?

caps = Gst::Caps.new("audio/x-raw")
caps["format"] = "F32LE"
caps["rate", :int] = 16_000
caps["channels", :int] = 1
caps["layout"] = "interleaved"

sink.caps = caps
sink.emit_signals = true

duration_threshold = 3_000_000_000 # nanoseconds
whisper = Whisper::Context.new("tiny-q8_0")
params = Whisper::Params.new(language: "ja")

buffer_list = Gst::BufferList.new

duration = 0
sink.signal_connect(:new_sample) do |_|
  sample = sink.pull_sample
  buffer = sample.buffer
  dur = buffer.duration
  next Gst::FlowReturn::OK if dur == Gst::CLOCK_TIME_NONE

  buffer_list.insert -1, sample.buffer
  duration += dur
  next Gst::FlowReturn::OK unless duration >= duration_threshold

  samples = Gst::Sample.new(Gst::Buffer.new, caps, Gst::Segment.new, caps.structures[0])
  samples.buffer_list = buffer_list
[...]

PR #1707, opened 2026-04-10
gstreamer: Make Gst::Buffer#map raise an exception on failure

KitaitiMakoto:
Hi,

This pull request makes `Gst::Buffer#map` raise an exception on failure as said at https://github.com/ruby-gnome/ruby-gnome/pull/1704#discussion_r3030949967

You know, this is a breaking change. Should we put a deprecation period that `Gst::Buffer#map` returns `[success, map]` or `[map, map]`?

Thank you.

### DIFF
--- a/gstreamer/lib/gst/buffer.rb
+++ b/gstreamer/lib/gst/buffer.rb
@@ -1,0 +1,39 @@
+# Copyright (C) 2026  Ruby-GNOME Project Team
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+module Gst
+  class Buffer
+    alias_method :map_raw, :map
+    def map(flags)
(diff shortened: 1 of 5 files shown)


gstreamer/lib/gst/buffer.rb
```diff
@@ -0,0 +1,33 @@
...
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+module Gst
+  class Buffer
+    alias_method :map_raw, :map
+    def map(flags)
+      success, info = map_raw(flags)
+      raise "failed to map buffer" unless success
```
kou: Let's use `Gst::CoreError::Failed` as error class:


```suggestion
      raise Gst::CoreError::Failed.new("failed to map buffer") unless success
```

gstreamer/lib/gst/buffer.rb
```diff
@@ -0,0 +1,33 @@
...
+      info
+    end
+
+    alias_method :map_range_raw, :map_range
+    def map_range(idx, length, flags)
+      success, info = map_range_raw(idx, length, flags)
+      raise "failed to map buffer range" unless success
```
kou: ditto.

gstreamer/lib/gst/loader.rb
```diff
@@ -61,6 +61,7 @@ def require_extension
     def require_libraries
       require "gst/bin"
       require "gst/bus"
+      require "gst/buffer"
```
kou: Could you keep this list in alphabetical order?

```suggestion
      require "gst/buffer"
      require "gst/bus"
```

kou (commented):
We don't need deprecation period. This is a trivial breaking change.

KitaitiMakoto:
Fixed.

I think failed test is a flaky test because the same code succeeded in my CI environment: https://github.com/KitaitiMakoto/ruby-gnome/actions/runs/24250070725/job/70806698882

kou (approved):
+1

It may be better that we also provide `buffer.map {}` for auto `unmap`.

PR #1708, opened 2026-04-12
gstreamer: Make Gst::Buffer#map accept block

KitaitiMakoto:
Hi,

I made `Gst::Buffer#map` accept block and unmap at the end of the block.

Thank you.


kou:
Thanks.

PR #1710, opened 2026-04-19
gstreamer: Don't raise exception in memory view function

KitaitiMakoto:
Hello,

I found out that exporting MemoryView potentially raises an exception and it leads resource leak, and fixed it.

Thank you.

### DIFF
--- a/gstreamer/ext/gstreamer/rbgst-sample.c
+++ b/gstreamer/ext/gstreamer/rbgst-sample.c
@@ -117,6 +117,8 @@ rg_gst_memory_view_init_from_audio(VALUE obj, rb_memory_view_t *view, int flags,
 
         return false;
     }
+    // Doesn't use ALLOC/ALLOC_N because they may raise an exception and then
+    // the map_info and audio_info won't be freed and the buffer won't be unrefed.
     map_info = g_new(GstMapInfo, 1);
     if (!gst_buffer_map(buffer, map_info, is_writable_requested ? GST_MAP_WRITE : GST_MAP_READ)) {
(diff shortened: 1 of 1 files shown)


gstreamer/ext/gstreamer/rbgst-sample.c
```diff
@@ -116,27 +117,18 @@ rg_gst_memory_view_init_from_audio(VALUE obj, rb_memory_view_t *view, int flags,
 
         return false;
     }
-    map_info = ALLOC(GstMapInfo);
+    map_info = g_new(GstMapInfo, 1);
     if (!gst_buffer_map(buffer, map_info, is_writable_requested ? GST_MAP_WRITE : GST_MAP_READ)) {
         rb_warn("Gst::Sample: failed to map buffer");
-        xfree(map_info);
+        g_free(map_info);
```
kou, line 125: Why do we need to use `g_new()`/`g_free()` instead of `ALLOC()`/`xfree()`?

KitaitiMakoto: `ALLOC` may raise an Ruby exception. The following resource managements like freeing `map_info`, `audio_info` and unrefing `buffer` will be skipped and leaked in that case.

kou: OK. Could you explain it in the PR description and/or the source comment?

KitaitiMakoto: Did by comments.

kou:
Thanks.

KitaitiMakoto:
Thank you for merge.

PR #1712, opened 2026-06-10
glib2: don't free a TypedData object's own rb_data_type_t in cinfo_free

rwstauner:
Each class registered through the rbgobj type system (G_DEF_CLASS,
GObjectIntrospection's define_class, ...) is wrapped by an
RGObjClassInfo TypedData object whose rb_data_type_t is heap allocated
per class in rbgobj_class_info_create_data_type. cinfo_free freed that
struct via xfree(cinfo->data_type).

That is a use-after-free. After invoking the dfree callback,
rb_data_free reads RTYPEDDATA_TYPE(obj)->flags to evaluate the
RUBY_TYPED_EMBEDDABLE check (Ruby >= 3.3, commit 392238e3fd "Implement
embedded TypedData objects"), so the type struct must outlive the free
callback. A subclass' rb_data_type_t may also reference it via ->parent.
AddressSanitizer reports a heap-use-after-free in rb_data_free during
rb_objspace_call_finalizer at interpreter shutdown, when these pinned
cinfo objects are finally collected.

Stop freeing cinfo->data_type. These per-class type descriptors live for
the lifetime of the registered class, so leaving them allocated is
intentional and matches the usual rb_data_type_t lifetime contract.

Add a regression test that loads glib2 in a child process and asserts it
exits cleanly; this fails (SIGABRT) under an ASAN/valgrind build before
the fix and passes after.

This essentially leaks the type.  A better alternative would be to declare the type statically as is usually done: https://github.com/ruby-gnome/ruby-gnome/pull/1714

### DIFF
--- a/glib2/ext/glib2/rbgobj_type.c
+++ b/glib2/ext/glib2/rbgobj_type.c
@@ -68,7 +68,13 @@ cinfo_free(void *data)
     RGObjClassInfo *cinfo = data;
     g_hash_table_remove(gtype_to_cinfo, GUINT_TO_POINTER(cinfo->gtype));
     xfree(cinfo->name);
-    xfree(cinfo->data_type);
+    /* Don't free cinfo->data_type here: it is this object's own
+     * rb_data_type_t, and the GC still reads its ->flags after this
+     * free callback returns (RUBY_TYPED_EMBEDDABLE check in
+     * rb_data_free, Ruby >= 3.3). Freeing it would be a use-after-free.
+     * A subclass' rb_data_type_t may also reference it via ->parent.
+     * These per-class type descriptors live for the lifetime of the
+     * registered class, so leaving them allocated is intentional. */
     xfree(cinfo);
 }
 
--- a/glib2/test/test-type-gc.rb
+++ b/glib2/test/test-type-gc.rb
@@ -1,0 +1,72 @@
+# Copyright (C) 2026  Ruby-GNOME Project Team
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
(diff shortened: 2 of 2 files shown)


glib2/ext/glib2/rbgobj_type.c
```diff
@@ -68,7 +68,13 @@ cinfo_free(void *data)
...
     g_hash_table_remove(gtype_to_cinfo, GUINT_TO_POINTER(cinfo->gtype));
     xfree(cinfo->name);
-    xfree(cinfo->data_type);
+    /* Don't free cinfo->data_type here: it is this object's own
+     * rb_data_type_t, and the GC still reads its ->flags after this
+     * free callback returns (RUBY_TYPED_EMBEDDABLE check in
+     * rb_data_free, Ruby >= 3.3). Freeing it would be a use-after-free.
+     * A subclass' rb_data_type_t may also reference it via ->parent.
+     * These per-class type descriptors live for the lifetime of the
+     * registered class, so leaving them allocated is intentional. */
```
kou, line 77: It seems that you fixed this in Ruby master: https://github.com/ruby/ruby/pull/17266

Let's do this only for Ruby < 4.1:

```ruby
#if RUBY_API_VERSION_CODE >= 40100
  /* ... */
  xfree(cinfo->data_type);
#endif
```

glib2/test/test-type-gc.rb
kou, line 1: We don't need this test. If ASAN is enabled, our test runner process is crashed, right? (BTW, how did you enable ASAN?)

rwstauner:
A better solution than this is to switch to a single statically declared type: https://github.com/ruby-gnome/ruby-gnome/pull/1714

glib2/test/test-type-gc.rb
rwstauner: I also added this test to the static type PR: https://github.com/ruby-gnome/ruby-gnome/pull/1714
I don't think that the test suite was crashing without the test, but it's hard to remember now.

rwstauner: When building ruby locally I used
```bash
    export cflags="-fsanitize=address -fsanitize-recover=address -fno-omit-frame-pointer -Wno-incompatible-function-pointer-types"
    export CXXFLAGS="-fsanitize=address -fsanitize-recover=address -fno-omit-frame-pointer"
```
then I just run this test suite with that ruby
```bash
make clean
ruby extconf.rb glib2
RUBY_GNOME_BUILD_DIR=$PWD ruby glib2/test/run-test.rb # -t TestGLibTypeGC
```

PR #1713, opened 2026-06-12
Remove vm_throw_data.reserved field when ruby >= 4.1

rwstauner:
The field was removed in https://github.com/ruby/ruby/commit/ff55280320


kou:
Oh... Let's reconsider the `break replaced` API... We don't want to maintain the API. (We don't want to follow the internal Ruby API change.)

PR #1715, opened 2026-06-14
glib2: Add missing close angle bracket in GLib::Param#inspect

mephistobooks:
Hi. i added close-angle-bracket in GLib::Param#inspect.

### DIFF
--- a/glib2/test/test-binding.rb
+++ b/glib2/test/test-binding.rb
@@ -171,17 +171,5 @@ class TestGLibBinding < Test::Unit::TestCase
       @source.source = 20
       assert_equal("30", @target.target)
     end
-
-    test "Param#inspect" do
-      tmp = GLib::Param::Int.new("source", # name
-                                 nil, # nick
-                                 nil, # blurb
-                                 0,     # min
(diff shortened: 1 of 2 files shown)


glib2/test/test-binding.rb
```diff
@@ -171,5 +171,17 @@ def target=(value)
...
+      tmp = GLib::Param::Int.new("source", # name
+                                 nil, # nick
+                                 nil, # blurb
+                                 0,     # min
+                                 100,   # max
+                                 0,     # default
+                                 [:readable, :writable])
+      assert_equal(GLib::Param::Int, tmp.class)
+      assert_match(/^#<.+>$/, tmp.inspect)
+    end
```
kou: We don't need this test.
Could you remove this?

kou (approved):
+1

Good catch!

glib2/test/test-binding.rb
mephistobooks: sure. I removed the test code.

kou:
Thanks.

PR #1724, opened 2026-07-16
WIP: migrate from girepository-1 to girepository-2

tvallois:
The scope of this pull request is to migrate to girepository-2 according to conversation with @kou [here](https://github.com/ruby-gnome/ruby-gnome/issues/1703#issuecomment-4182728842)


tvallois:
@kou , i'm struggling a bit to continue the migration from what i've done. If i'm understanding the code correctly (sorry first time i'm looking at internals in Ruby, GLib, etc.), girepository-2 does not accept overload ruby-gnome do with methods such as:
```ruby
GType
gi_constant_info_get_type(void)
{
    static GType type = 0;
    if (type == 0) {
	    type = g_boxed_type_register_static("GIConstantInfo",
                                            (GBoxedCopyFunc)gi_base_info_ref,
                                            (GBoxedFreeFunc)gi_base_info_unref);
    }
    return type;
}
```

This [function](https://github.com/GNOME/glib/blob/main/girepository/gibaseinfo.c#L274) raise an error. When i remove those methods in ruby-gnome, it breaks `rb_gi_base_info_to_ruby` function because gi types are not boxed types anymore. I'm kinda lost at the solution i should follow. Can you advise me a little bit? (maybe i'm in the wrong way as well :D)

kou:
Thanks for working on this!

We don't need these type definitions in Ruby GNOME if girepository-2 provides them. We defined them because girepository-1 doesn't provide them.

BTW, I want to mix the girepository-2 bindings into `glib2/ext/glib2/` not a separated `glib2/ext/gobject-introspection/. Can I create a base part as the first step?

tvallois:
> Thanks for working on this!
> 
> We don't need these type definitions in Ruby GNOME if girepository-2 provides them. We defined them because girepository-1 doesn't provide them.
> 
> BTW, I want to mix the girepository-2 bindings into `glib2/ext/glib2/` not a separated `glib2/ext/gobject-introspection/. Can I create a base part as the first step?

Yeah no problem, how do you want to proceed? In a separate branch?

tvallois:
I've moved gobject-introspection into glib2. Tests are not passing anymore with `ruby test/run-test.rb test-repository`. If you have any clue what's going on... :)

kou:
Oh sorry. I haven't worked on this yet... Please wait for a few days...

kou:
I tried this and noticed that we need to improve `glib2/ext/glib2/rbgobject.c` to integrate with girepository-2.0 because `GIBaseInfo` is a new fundamental `GType`.

I'm working on this. Please wait for a few more days...

kou:
#1729 is the minimum first step.

PR #1727, opened 2026-07-25
glib2: Destroy notify queue

KitaitiMakoto:
Hello,

This pull request adds a queue for running `GDestroyNotify` callbacks in a Ruby thread. See comments in glib2/ext/glib2/rbgutil_destroy_notify.c for the design.

I implemented this feature to make GStreamer's `Gst::Sample` a MemoryView consumer. We need to release MemoryViews when `Gst::Buffer`s included in `Gst::Sample`s are freed. But, callbacks invoked at that time run in a GStreamer thread where Ruby APIs are not available. To call Ruby APIs such as `rb_memory_view_release()`, we need to dispatch callbacks to a Ruby thread.

In this case, MemoryView release callbacks may occur frequently. So, synchronous `callback_request_queue` would block a Ruby thread too frequently. That's why I added another queue feature.

Can you review this?

Thank you.

### DIFF
--- a/glib2/ext/glib2/glib2.def
+++ b/glib2/ext/glib2/glib2.def
@@ -162,6 +162,8 @@ EXPORTS
 	rbgutil_invoke_callback
 	rbgutil_start_callback_dispatch_thread
 	rbgutil_stop_callback_dispatch_thread
+	rbgutil_queue_destroy_notify_request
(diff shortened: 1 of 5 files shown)


kou:
How about reusing the existing queue like #1731?

KitaitiMakoto:
The idea seems good! I will try the code.

PR #1728, opened 2026-08-07
gstreamer: Make Sample from audio data MemoryView

KitaitiMakoto:
Hi,

I added a method `GstAudio.audio_make_sample` to make `Gst::Sample`s from audio data MemoryView producers.

~This pull request is based on #1727.~
Rebased on #1731, which was merged into main.

Currently, the feature is enabled when Ruby >= 4.1.0 because currently released versions contains a bug fixed by https://github.com/ruby/ruby/pull/17851.

I hope the fix will be backported to 4.0.7. Until the version is released, I will keep this pull request as a draft and wait to see whether it will be backported.


KitaitiMakoto:
Ruby 4.0.7 doesn't include the fix. We seem to need to wait for the next Christmas.

PR #1731, opened 2026-08-16
glib2: add rbgutil_invoke_callback_async()

kou:
GitHub: Fix GH-1727

It's a fire-and-forget variant of rbgutil_invoke_callback() for callbacks that are invoked frequently from non-Ruby threads and don't need to report a result, such as GDestroyNotify for rb_memory_view_release().

rbgutil_invoke_callback() blocks the calling thread until a newly created Ruby thread finishes running the callback. It's too heavy for frequent callbacks such as releasing a memory view on GstBuffer free in a GStreamer streaming thread.

This reuses the existing callback dispatch infrastructure (callback_request_queue and the callback dispatch thread started by rbgutil_start_callback_dispatch_thread()) instead of adding a new queue. An async request is heap-allocated, pushed to the same queue and run in the dispatch thread directly without creating a Ruby thread per request.

Reported by KITAITI Makoto. Thanks!!!

### DIFF
--- a/glib2/ext/glib2/glib2.def
+++ b/glib2/ext/glib2/glib2.def
@@ -160,6 +160,7 @@ EXPORTS
 	rbgutil_protect
 	rbgutil_on_callback_error
 	rbgutil_invoke_callback
+	rbgutil_invoke_callback_async
 	rbgutil_start_callback_dispatch_thread
 	rbgutil_stop_callback_dispatch_thread
 	rbgutil_string_set_utf8_encoding
--- a/glib2/ext/glib2/rbgutil.h
+++ b/glib2/ext/glib2/rbgutil.h
@@ -117,6 +117,7 @@ extern void rbgutil_set_properties(VALUE self, VALUE hash);
 extern VALUE rbgutil_protect(VALUE (*proc) (VALUE), VALUE data);
 extern void rbgutil_on_callback_error(VALUE error);
 extern VALUE rbgutil_invoke_callback(VALUE (*func)(VALUE), VALUE arg);
+extern void rbgutil_invoke_callback_async(VALUE (*func)(VALUE), VALUE arg);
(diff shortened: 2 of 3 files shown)


KitaitiMakoto:
This patch is simple and clean, and works fine for my simple script just to write audio MemoryView producers to a file.

One thing: is adding the function name to `gobject-introspection/ext/gobject-introspection/extconf.rb` intended to another pull request?

I encountered SEGV in test of #1728. It occurs after test ends. I guess test-unit architecture and Ruby VM termination affects the problem. Anyway, that problem exists ever (even for synchronous version) and is not related to this pull request. I will make a minimal reproducible code and file an issue.

Thanks for your work!

kou:
> One thing: is adding the function name to `gobject-introspection/ext/gobject-introspection/extconf.rb` intended to another pull request?

I think that we don't need to add anything to `gobject-introspection/` with this approach. What is needed?

KitaitiMakoto:
I just thought it is needed because the file has `"_rbgutil_invoke_callback"`. But I don't understand what the file do. I don't object at all if you think it's okay.

Without adding the name, my patch and script work without any trouble.

kou:
Ah, I understand. `_rbgutil_invoke_callback` is needed because `rbgutil_invoke_callback()` is used in `gobject-introspection/`. `_rbgutil_invoke_callback_async` isn't needed because `rbgutil_invoke_callback_async()` isn't used in `gobject_introspection/`.

KitaitiMakoto:
Ah, I confirmed `rbgutil_invoke_callback` used in GObject Introspection. Thank you for your explanation.

KitaitiMakoto:
I found when I reduce a reference count of an object to negative number, segmentation faults. I will investigate needles reducing of the reference count in my test.

PR #1741, opened 2026-08-22
Reproducible patch for causing SEGV by callback

KitaitiMakoto:
Hello,

This is a minimal reproducible patch for the segmentation fault in GLib2's callback feature I mentioned at https://github.com/ruby-gnome/ruby-gnome/pull/1731#issuecomment-5309756423.

I should see the SEGV by running:

    % cd glib2
    % ruby ./test/run-test.rb test-callback

This is not intended to be merged. Just for reporting.

I will keep researching the fix for this problem.

Thanks.


KitaitiMakoto:
This is wrong code. The cause was found and fixed at #1742.

